PR #26, opened 2024-11-04
Use spanvalue to format values

apstndb:
* Replace most logic in `decoder.go` with [apstndb/spanvalue](https://github.com/apstndb/spanvalue).
* Set `DisableNativeMetrics: true` as workaround https://github.com/googleapis/google-cloud-go/issues/11080


apstndb:
https://github.com/apstndb/spanner-mycli/pull/26/commits/00883b6d8dc9bb85c3ca6cc60873974e90f042f3 is needed to suppress:
```
google: could not find default credentials. See https://cloud.google.com/docs/authentication/external/set-up-adc for more information
```

PR #296, opened 2025-06-17
Fix unintended changes in --help output and README.md inconsistencies

### LINKED ISSUE #285: Fix unintended changes in --help output and README.md inconsistencies
### 1. Unintended default value display in --set flag help text

The `--set` flag help text is showing an unintended default value display:

```
--set=                                              Set system variables e.g. --set=name1=value1 --set=name2=value2 (default: {CLI_PARSE_MODE:FALLBACK})
```

This appears to be caused by the go-flags library displaying struct field values in help text, even when they shouldn't be shown for map-type flags.

### 2. README.md help output inconsistency

The help output in README.md is inconsistent with the actual `--help` command output:

**README.md (line 98):**
```
Usage:
  spanner-mycli [OPTIONS]
```

**Actual `go run . --help` output:**
```
Usage:
  spanner-mycli
```

This inconsistency was discovered when implementing the new help output update workflow with `scripts/docs/update-help-output.sh`.

### 1. --set flag help text should only show:
```
--set=                                              Set system variables e.g. --set=name1=value1 --set=name2=value2
```

### 2. README.md should match actual help output:
```
Usage:
  spanner-mycli
```

### Phase 1: Fix --set flag help text
- Apply `default-mask:"-"` struct tag to --set flag field (following CLAUDE.md Issue #251 insights)
- Verify other map-type flags don't have similar issues

### Phase 2: Update README.md help sections
- Use `scripts/docs/update-help-output.sh` to generate current help output
- Replace README.md help sections (lines 97-135 and 355-406) with generated content
- Verify generation comments are updated to reference the new script

### Phase 3: Establish consistency checks
- Document the help output update workflow
- Ensure README.md generation comments reference `scripts/docs/update-help-output.sh`

## Success Criteria

- [ ] --set flag help text shows no unintended default values
- [ ] README.md help output matches actual `--help` command output
- [ ] README.md statement help table matches actual `--statement-help` output
- [ ] Generation comments in README.md reference correct update script
- [ ] No other flags show unintended default value displays

## Related

This combines two help-related issues discovered during:
- Implementation of --timeout flag (Issue #276)
- Documentation splitting and help workflow automation (current PR)
- go-flags library behavior documented in CLAUDE.md Issue #251 insights

apstndb:
## Summary

This PR fixes help output inconsistencies identified in issue #285 by implementing comprehensive default-mask handling and restoring proper help flag behavior.

### 1. Restore [OPTIONS] in help output
- Enable `flags.HelpFlag` in the parser configuration
- Remove redundant manual help flag handling

### 2. Fix --set default display issue
- Apply `default-mask:"-"` comprehensively to all `spannerOptions` fields
- Prevent config file values from appearing as misleading defaults in help output

### 3. Update README.md
- Add missing "Help Options" section to match actual help output

### 4. Simplify struct documentation
- Streamline comments explaining go-flags behavior with default-mask

### go-flags Default Value Display Behavior

The core issue was that `go-flags` displays **any non-zero struct field values** as defaults in help text, not just values from `default` tags. This behavior becomes problematic with spanner-mycli's two-parser architecture:

1. **Config file parser** (`flags.Default`) - loads `.spanner_mycli.cnf` and sets struct field values
2. **Command line parser** (`flags.HelpFlag`) - processes CLI args and generates help output

Without `default-mask:"-"`, help output would show config file values as defaults, creating inconsistent and misleading help text that varies based on user's config file contents.

### Comprehensive default-mask Application

Applied `default-mask:"-"` to **all** `spannerOptions` fields for consistency, including:
- Visible config-overridable fields (prevents config value leakage)
- Hidden fields (`hidden:"true"`) - for consistency, even though hidden fields never appear in help output
- Fields with meaningful defaults use `default-mask:"actual-value"` (e.g., `Prompt`, `VertexAIModel`)

### Key Discovery: go-flags hidden Field Behavior
Through testing, we confirmed that **fields with `hidden:"true"` are completely excluded from help output**, regardless of their values. Even when config file values are loaded into hidden fields, they never appear in help text.

This means:
- `hidden:"true"` alone is sufficient to exclude fields from help
- Adding `default-mask:"-"` to hidden fields is technically unnecessary but improves consistency

### Testing Approach
Created test config file with values for both visible and hidden fields:
```ini
[spanner]
priority = HIGH
sql = "hidden-field-test"  # This never appears in help due to hidden:"true"
```

Confirmed that:
- Hidden fields remain invisible in help output regardless of config values
- Visible fields without `default-mask` would show config values as defaults
- Fields with `default-mask:"-"` properly hide current values from help display

### Architecture Pattern
This establishes a clear pattern for future CLI option additions:
- **All fields**: Use `default-mask:"-"` for consistency
- **Exception**: Fields with meaningful display defaults use `default-mask:"value"`
- **Simple rule**: No need to distinguish between config-overridable vs non-config-overridable fields

## Verification

All changes maintain backward compatibility while fixing the help output inconsistencies described in issue #285.

Fixes #285

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,6 +134,9 @@ spanner:
       --enable-partitioned-dml                            Partitioned DML as default (AUTOCOMMIT_DML_MODE=PARTITIONED_NON_ATOMIC)
       --timeout=                                          Statement timeout (e.g., '10s', '5m', '1h') (default: 10m)
       --mcp                                               Run as MCP server
+
+Help Options:
+  -h, --help                                              Show this help message
 ```
 
 ### Authentication
--- a/main.go
+++ b/main.go
@@ -54,41 +54,49 @@ type globalOptions struct {
 }
 
 // We can't use `default` because spanner-mycli uses multiple flags.NewParser() to process config files and flags.
+// 
+// Config-overridable fields use `default-mask:"-"` to prevent config file values from appearing in help output.
+// This is necessary because:
+// 1. spanner-mycli uses a two-parser approach: config file parser → command line parser
+// 2. go-flags displays struct field values as defaults in help text (not just `default` tags)
+// 3. Config file parsing sets field values before help generation, causing config values to appear as defaults
+// 4. Users would see misleading defaults like "(default: test-value)" when they have values in config files
+// 5. Without default-mask, help output changes based on user's config file contents, creating inconsistent UX
 type spannerOptions struct {
 	ProjectId                 string            `long:"project" short:"p" env:"SPANNER_PROJECT_ID" default-mask:"-" description:"(required) GCP Project ID."`
 	InstanceId                string            `long:"instance" short:"i" env:"SPANNER_INSTANCE_ID" default-mask:"-" description:"(required) Cloud Spanner Instance ID"`
 	DatabaseId                string            `long:"database" short:"d" env:"SPANNER_DATABASE_ID" default-mask:"-" description:"Cloud Spanner Database ID. Optional when --detached is used."`
 	Detached                  bool              `long:"detached" description:"Start in detached mode, ignoring database env var/flag"`
-	Execute                   string            `long:"execute" short:"e" description:"Execute SQL statement and quit. --sql is an alias."`
-	File                      string            `long:"file" short:"f" description:"Execute SQL statement from file and quit."`
+	Execute                   string            `long:"execute" short:"e" description:"Execute SQL statement and quit. --sql is an alias." default-mask:"-"`
+	File                      string            `long:"file" short:"f" description:"Execute SQL statement from file and quit." default-mask:"-"`
 	Table                     bool              `long:"table" short:"t" description:"Display output in table format for batch mode."`
 	Verbose                   bool              `long:"verbose" short:"v" description:"Display verbose output."`
-	Credential                string            `long:"credential" description:"Use the specific credential file"`
+	Credential                string            `long:"credential" description:"Use the specific credential file" default-mask:"-"`
 	Prompt                    *string           `long:"prompt" description:"Set the prompt to the specified format" default-mask:"spanner%t> "`
 	Prompt2                   *string           `long:"prompt2" description:"Set the prompt2 to the specified format" default-mask:"%P%R> "`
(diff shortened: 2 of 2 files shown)


apstndb:
Thank you for the review\! 

Regarding the CLAUDE.md Update Rules section:

The addition of these rules is intentional and necessary. This section serves as self-documenting guidelines for maintaining CLAUDE.md's focus and preventing it from becoming bloated over time. Without these explicit rules, developers (both human and AI) might continue adding verbose content that would reduce the effectiveness of this file.

The rules section itself is concise and directly addresses a critical maintenance concern - ensuring CLAUDE.md remains focused on essential information that would cause problems if skipped, rather than comprehensive documentation that belongs in specialized files.

This follows the principle stated in the rules: "CLAUDE.md should only contain information that would cause problems if skipped" - and these rules themselves meet that criteria since ignoring them could lead to documentation bloat issues.

The default-mask comment has already been implemented as suggested in the code.

apstndb:
## Additional Fix: Comprehensive default-mask Application

I've added the missing  to all config-overridable fields as discussed. This was necessary to prevent config file values from leaking into help output.

### Problem Demonstration
With a config file containing:
```ini
priority = "HIGH"
role = "test-role"
endpoint = "test-endpoint:443"
```

**Before the fix**, help output showed misleading defaults:
```
--priority=    (default: HIGH)
--role=        (default: test-role)  
--endpoint=    (default: test-endpoint:443)
```

**After the fix**, these fields show no defaults (correct behavior):
```
--priority=    Set default request priority (HIGH|MEDIUM|LOW)
--role=        Use the specific database role
--endpoint=    Set the Spanner API endpoint (host:port)
```

### Why This Is Critical
1. **go-flags behavior**: Displays struct field values as defaults in help text, not just `default` tags
2. **Two-parser architecture**: Config file parser → command line parser means config values are loaded before help generation
3. **User confusion**: Without default-mask, help output changes based on user's config file contents
4. **Inconsistent UX**: Different users see different "defaults" based on their config files

### Technical Details
The fix adds comprehensive documentation in the struct comment explaining why this is necessary and applies `default-mask:"-"` to all string/map fields that can be set via config files, preventing value leakage while preserving legitimate defaults like `prompt`, `history`, `emulator-image`, etc.

PR #382, opened 2025-07-01
feat: implement --skip-column-names option

### LINKED ISSUE #351: Implement --skip-column-names option
## Overview

Implement the `--skip-column-names` option to suppress column headers in output, compatible with Google Cloud Spanner CLI.

## Specification

- Flag: `--skip-column-names`
- Suppresses column headers in query result output
- Useful for scripting and data processing

## Implementation notes

- Should work with table format output
- Consider interaction with other output formats when implemented
- Affects only the header row, not the data rows

#### Default behavior (with headers):
```bash
$ spanner-mycli -e "SELECT 1 AS foo, 2 AS bar"
+-----+-----+
| foo | bar |
+-----+-----+
| 1   | 2   |
+-----+-----+
```

#### With --skip-column-names (table format):
```bash
$ spanner-mycli --skip-column-names -e "SELECT 1 AS foo, 2 AS bar"
+---+---+
| 1 | 2 |
+---+---+
```

#### With --skip-column-names (tab format):
```bash
$ spanner-mycli --skip-column-names --table -e "SELECT 1 AS foo, 2 AS bar"
1	2
```

#### Using system variable:
```bash
$ spanner-mycli -e "SET CLI_SKIP_COLUMN_NAMES = TRUE; SELECT 1 AS foo, 2 AS bar"
+---+---+
| 1 | 2 |
+---+---+

# Or with tab format
$ spanner-mycli -e "SET CLI_FORMAT = 'TAB'; SET CLI_SKIP_COLUMN_NAMES = TRUE; SELECT 1 AS foo, 2 AS bar"
1	2
```

#### Vertical format (headers preserved):
```bash
$ spanner-mycli --skip-column-names -e "SET CLI_FORMAT = 'VERTICAL'; SELECT 1 AS foo, 2 AS bar"
*************************** 1. row ***************************
foo: 1
bar: 2
```

## Priority

Medium priority - useful for scripting and automation

## References

- Parent issue: #342
- [Google Cloud Spanner CLI documentation](https://cloud.google.com/spanner/docs/spanner-cli)
EOF < /dev/null

apstndb:
## Summary
- Implements the `--skip-column-names` flag to suppress column headers in output
- Adds corresponding `CLI_SKIP_COLUMN_NAMES` system variable for runtime control
- Compatible with Google Cloud Spanner CLI

## Details

This PR implements the `--skip-column-names` option as specified in #351. The implementation allows users to suppress column headers in query output, which is useful for scripting and data processing.

### Implementation
- Added `SkipColumnNames` field to `systemVariables` struct
- Added `--skip-column-names` CLI flag to `spannerOptions` 
- Modified `printTableData()` to conditionally skip headers based on the flag
- Headers are suppressed for table and tab formats, preserved for vertical format

### Usage Examples

```bash
# Suppress headers with flag
spanner-mycli --skip-column-names -e "SELECT 1 AS foo, 2 AS bar"
+---+---+
| 1 | 2 |
+---+---+

# Suppress headers with system variable
spanner-mycli -e "SET CLI_SKIP_COLUMN_NAMES = TRUE; SELECT 1 AS foo, 2 AS bar"
+---+---+
| 1 | 2 |
+---+---+

# Tab format without headers
spanner-mycli --skip-column-names --table -e "SELECT 1 AS foo, 2 AS bar"
1	2
```

## Test plan
- [x] Unit tests for system variable (system_variables_test.go)
- [x] Integration tests for table and tab formats (cli_test.go)
- [x] Manual testing with different output formats
- [x] `make check` passes

Fixes #351

🤖 Generated with [Claude Code](https://claude.ai/code)

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -141,6 +141,7 @@ spanner:
       --try-partition-query                               Test whether the query can be executed as partition query without execution
       --mcp                                               Run as MCP server
       --skip-system-command                               Do not allow system commands
+      --skip-column-names                                 Suppress column headers in output
 
 Help Options:
   -h, --help                                              Show this help message
--- a/cli_output.go
+++ b/cli_output.go
@@ -87,7 +87,9 @@ func printTableData(sysVars *systemVariables, screenWidth int, out io.Writer, re
 				slices.Values(adjustedWidths))),
 		)
 
-		table.Header(headers)
+		if !sysVars.SkipColumnNames {
+			table.Header(headers)
+		}
 
 		for _, row := range rows {
 			wrappedColumns := slices.Collect(hiter.Unify(
@@ -137,7 +139,9 @@ func printTableData(sysVars *systemVariables, screenWidth int, out io.Writer, re
 		}
 	case DisplayModeTab:
 		if len(columnNames) > 0 {
-			fmt.Fprintln(out, strings.Join(columnNames, "\t"))
+			if !sysVars.SkipColumnNames {
+				fmt.Fprintln(out, strings.Join(columnNames, "\t"))
+			}
 			for _, row := range result.Rows {
 				fmt.Fprintln(out, strings.Join(row, "\t"))
 			}
--- a/cli_test.go
+++ b/cli_test.go
@@ -394,6 +394,52 @@ bar: 4
 			t.Errorf("invalid print: expected = %s, but got = %s", expected, got)
 		}
 	})
+
+	t.Run("SkipColumnNames with DisplayModeTable", func(t *testing.T) {
+		out := &bytes.Buffer{}
+		result := &Result{
+			TableHeader: toTableHeader("foo", "bar"),
+			Rows: sliceOf(
+				toRow("1", "2"),
+				toRow("3", "4"),
+			),
+			IsMutation: false,
+		}
+		printResult(&systemVariables{CLIFormat: DisplayModeTable, SkipColumnNames: true}, math.MaxInt, out, result, false, "")
+
+		expected := strings.TrimPrefix(`
++---+---+
+| 1 | 2 |
+| 3 | 4 |
++---+---+
+`, "\n")
+
+		got := out.String()
+		if got != expected {
+			t.Errorf("invalid print: expected = %s, but got = %s", expected, got)
+		}
+	})
+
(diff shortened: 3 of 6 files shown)


apstndb:
## Documentation Notes

After reviewing the implementation, I've identified some additional information that would be helpful to document:

### 1. Behavior with Different Output Formats

The implementation has specific behavior for each output format that should be clearly documented:

- **Table format (`CLI_FORMAT='TABLE'`)**: Headers are suppressed when `--skip-column-names` is set
- **Tab format (`CLI_FORMAT='TAB'`)**: Headers are suppressed when `--skip-column-names` is set
- **Vertical format (`CLI_FORMAT='VERTICAL'`)**: Headers are **always preserved** regardless of the flag, as they are integral to the vertical format structure
- **Table comment formats (`CLI_FORMAT='TABLE_COMMENT'` and `CLI_FORMAT='TABLE_DETAIL_COMMENT'`)**: Headers follow the same suppression rules as table format

### 2. Interaction with Verbose Output

When `--verbose` or `CLI_VERBOSE=TRUE` is set, the system will still render tables even with empty result sets. In this case:
- With `--skip-column-names`: Only the table borders are shown (no header row)
- Without `--skip-column-names`: The header row is shown even for empty results

### 3. System Variable Precedence

The `CLI_SKIP_COLUMN_NAMES` system variable can be set at runtime and will override the initial `--skip-column-names` flag value:

```sql
-- Start with flag: spanner-mycli --skip-column-names
SET CLI_SKIP_COLUMN_NAMES = FALSE;  -- Re-enables headers
SELECT * FROM users;  -- Will show headers
```

### 4. Compatibility Note

This feature is designed to be compatible with Google Cloud Spanner CLI's `--skip-column-names` option, providing consistent behavior for users migrating from or using both tools interchangeably.

### 5. Use Cases

Primary use cases for this feature:
- **Shell scripting**: When piping output to other commands that expect raw data
- **Data export**: When generating CSV/TSV files without headers
- **Automated processing**: When the output is consumed by programs that parse data rows directly

These details could be added to the relevant documentation (README.md or system_variables.md) in a future update if needed.
EOF < /dev/null

PR #648, opened 2026-06-04
Stream non-table output by default

apstndb:
## Summary

- Make `CLI_STREAMING` control only table layout streaming.
- Stream non-table query formats even when table streaming is disabled.
- Remove the old buffered formatter registry and route SQL export/DUMP fallback through the streaming SQL writer.
- Keep DML THEN RETURN output atomic by returning pre-rendered output instead of exposing returned rows.
- Update SQL export, DUMP, transaction, and formatter tests for the new output paths.

## Verification

- `make check`
- Fetched `origin/main`; this branch is 3 commits ahead and 0 commits behind.

### DIFF
--- a/internal/mycli/execute_dml.go
+++ b/internal/mycli/execute_dml.go
@@ -100,6 +100,8 @@ func executeDML(ctx context.Context, session *Session, sql string) (*Result, err
 		queryStats = updateResult.Stats
 		tableHeader = toTableHeader(updateResult.Metadata.GetRowType().GetFields())
 		if tableHeader != nil {
+			// Render inside the transaction callback so a formatting error
+			// aborts the implicit commit instead of committing without output.
 			renderedOutput, err = renderDMLReturnedRows(session.systemVariables, tableHeader, updateResult.Rows)
 			if err != nil {
 				return 0, nil, nil, err
--- a/internal/mycli/execute_dml.go
+++ b/internal/mycli/execute_dml.go
@@ -1,6 +1,7 @@
 package mycli
 
 import (
+	"bytes"
 	"context"
 	"errors"
 	"slices"
@@ -87,15 +88,24 @@ func executeDML(ctx context.Context, session *Session, sql string) (*Result, err
 		return nil, err
 	}
 
-	var rows []Row
+	var renderedOutput []byte
+	var hasRenderedOutput bool
 	var queryStats map[string]any
+	var tableHeader TableHeader
 	result, err := session.RunInNewOrExistRwTx(ctx, func(tx *spanner.ReadWriteStmtBasedTransaction, implicit bool) (affected int64, plan *sppb.QueryPlan, metadata *sppb.ResultSetMetadata, err error) {
 		updateResult, err := session.runUpdateOnTransaction(ctx, tx, stmt, implicit)
 		if err != nil {
 			return 0, nil, nil, err
 		}
-		rows = updateResult.Rows
 		queryStats = updateResult.Stats
+		tableHeader = toTableHeader(updateResult.Metadata.GetRowType().GetFields())
+		if tableHeader != nil {
+			renderedOutput, err = renderDMLReturnedRows(session.systemVariables, tableHeader, updateResult.Rows)
+			if err != nil {
+				return 0, nil, nil, err
+			}
+			hasRenderedOutput = true
+		}
 		return updateResult.Count, updateResult.Plan, updateResult.Metadata, nil
 	})
 	if err != nil {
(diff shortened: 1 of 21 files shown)


internal/mycli/execute_dml.go
```diff
@@ -87,15 +88,24 @@ func executeDML(ctx context.Context, session *Session, sql string) (*Result, err
...
 	result, err := session.RunInNewOrExistRwTx(ctx, func(tx *spanner.ReadWriteStmtBasedTransaction, implicit bool) (affected int64, plan *sppb.QueryPlan, metadata *sppb.ResultSetMetadata, err error) {
 		updateResult, err := session.runUpdateOnTransaction(ctx, tx, stmt, implicit)
 		if err != nil {
 			return 0, nil, nil, err
 		}
-		rows = updateResult.Rows
 		queryStats = updateResult.Stats
+		tableHeader = toTableHeader(updateResult.Metadata.GetRowType().GetFields())
+		if tableHeader != nil {
+			renderedOutput, err = renderDMLReturnedRows(session.systemVariables, tableHeader, updateResult.Rows)
```
apstndb, line 105: **suggestion (document the load-bearing placement):** Rendering `THEN RETURN` output here — inside the `RunInNewOrExistRwTx` closure, before the implicit commit — is intentional but non-obvious. `RunInNewOrExistRwTxLocked` runs the closure once and commits an implicit transaction only *after* `f` returns, so a formatting error returned from `renderDMLReturnedRows` rolls back the DML. That keeps the side effect and the rendered output atomic (either both happen or neither), which is the whole reason to pre-render instead of returning `Rows`.

The trade-off is that the implicit transaction stays open for the duration of formatting. That's acceptable, but a one-line comment stating *why* the render is inside the closure would stop a future reader (or an automated reviewer) from "simplifying" it by moving the render after commit — which would silently change the failure mode to "DML committed but output errored."

internal/mycli/execute_sql.go
```diff
@@ -355,41 +367,27 @@ func executeStreamingSQL(ctx context.Context, qe *queryExecution) (*Result, erro
...
-		// Unknown mode: buffer for safety
-		return nil, nil
 	}
 
-	if !shouldStream {
+	switch sysVars.Query.StreamingMode {
+	case enums.StreamingModeTrue, enums.StreamingModeFalse, enums.StreamingModeAuto:
+	default:
 		return nil, nil
 	}
```
apstndb, line 391: **simplification:** This `switch` matches the valid modes with an empty body and then falls through to the `return createStreamingProcessorForMode(...)` below, so at a glance it reads like a no-op. The actual rule — "non-table formats stream for any valid `CLI_STREAMING` value; only an unexpected enum value buffers" — is easy to miss. Returning from inside the switch makes the intent explicit and drops the trailing return:

```go
// Non-table formats always stream regardless of CLI_STREAMING;
// only guard against an unexpected enum value.
switch sysVars.Query.StreamingMode {
case enums.StreamingModeTrue, enums.StreamingModeFalse, enums.StreamingModeAuto:
    return createStreamingProcessorForMode(sysVars.Display.CLIFormat, out, sysVars, screenWidth)
default:
    return nil, nil
}
```

Minor: in the table-mode branch above, the `case StreamingModeFalse, StreamingModeAuto:` and `default:` arms both `return nil, nil`, so they could collapse into a single `default` (keeping the "buffer for accurate column widths" comment).

internal/mycli/statements_dump.go
```diff
@@ -250,6 +240,19 @@ func writeResultRows(out io.Writer, rows []Row) error {
...
+func executeDumpTableBuffered(ctx context.Context, session *Session, txn *spanner.ReadOnlyTransaction, selectQuery, table string) (*Result, string, error) {
+	var buf bytes.Buffer
+	originalStreamManager := session.systemVariables.StreamManager
+	session.systemVariables.StreamManager = streamio.NewStreamManager(nil, &buf, &buf)
+	defer func() {
+		session.systemVariables.StreamManager = originalStreamManager
+	}()
+
+	result, err := executeSQLWithFormatAndTxn(ctx, session, txn, selectQuery,
+		enums.DisplayModeSQLInsert, enums.StreamingModeFalse, table)
```
apstndb: **clarity/nit:** After this PR, SQL export is a non-table format and therefore always streams regardless of `CLI_STREAMING`. So the `enums.StreamingModeFalse` passed here no longer forces buffering — buffering is achieved entirely by the `StreamManager` swap above that redirects output into `buf`. The code is correct, but the `StreamingModeFalse` argument (and the `...Buffered` name) now read as if they drive the buffering, which they don't. A one-line comment noting that buffering here comes from the `StreamManager` swap, not the streaming mode, would keep a future reader from relying on `StreamingModeFalse` to buffer.

apstndb (commented):
Reviewed the branch diff for correctness and simplification. No correctness defects found — this is a clean refactor (net −616 lines, CI green). I checked the spots most likely to regress and they hold up:

- `toTableHeader(GetFields())` returns `nil` for empty/nil row types, so plain DML (no `THEN RETURN`) does not set `HasRenderedOutput` and still prints nothing.
- The width baked into the DML pre-render (`displayScreenWidth` → `StreamManager.GetTerminalWidth`) resolves to the same terminal as `printResult`'s `size` (`GetTerminalSizeWithTty`), since both prefer the shared `ttyStream` and fall back to the same output writer — no width skew between DML `THEN RETURN` and other buffered tables.
- Dropping `screenWidth` on the new non-table buffered path is safe: the old non-table formatters ignored it.

Three optional clarity/simplification notes inline; none are blocking.

internal/mycli/execute_dml.go
apstndb: Addressed in ff9bc0f: added an inline comment explaining that DML THEN RETURN rendering stays inside the transaction callback to preserve commit/output atomicity.

internal/mycli/statements_dump.go
apstndb: Addressed in ff9bc0f: renamed the DUMP helper and documented that buffering comes from capturing the temporary StreamManager output.

internal/mycli/statements_dump.go
```diff
@@ -250,6 +240,21 @@ func writeResultRows(out io.Writer, rows []Row) error {
 	return nil
 }
 
+func executeDumpTableIntoBuffer(ctx context.Context, session *Session, txn *spanner.ReadOnlyTransaction, selectQuery, table string) (*Result, string, error) {
+	var buf bytes.Buffer
+	originalStreamManager := session.systemVariables.StreamManager
+	session.systemVariables.StreamManager = streamio.NewStreamManager(nil, &buf, &buf)
```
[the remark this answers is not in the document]
apstndb: Addressed in 3c79d33: the temporary DUMP StreamManager now keeps the original error stream, so warnings or errors cannot be written into the SQL dump buffer.

PR #654, opened 2026-06-04
Use spanvalue protofmt

apstndb:
## Summary

- update `github.com/apstndb/spanvalue` from `v0.5.0` to `v0.5.1`
- replace spanner-mycli's local PROTO/ENUM display plugins with `spanvalue/protofmt`
- remove local descriptor resolver construction, base64 decoding, proto unmarshalling, and enum lookup code now covered by spanvalue

## Verification

- `go test ./internal/mycli/decoder`
- `make check`

### DIFF
--- a/internal/mycli/decoder/decoder.go
+++ b/internal/mycli/decoder/decoder.go
@@ -35,6 +35,8 @@ func DecodeRow(row *spanner.Row) ([]string, error) {
 // If fds is nil, it returns a config without custom proto/enum support.
 // The multiline parameter controls the formatting of protobuf messages.
 func FormatConfigWithProto(fds *descriptorpb.FileDescriptorSet, multiline bool) (*spanvalue.FormatConfig, error) {
+	// protofmt preserves the old nil descriptor behavior: nil builds an empty
+	// resolver, so PROTO/ENUM plugins fall through to the default formatter.
 	resolver, err := protofmt.ProtoEnumResolverFromFileDescriptorSet(fds)
 	if err != nil {
 		return nil, err
--- a/internal/mycli/decoder/decoder.go
+++ b/internal/mycli/decoder/decoder.go
@@ -17,23 +17,13 @@
 package decoder
 
 import (
-	"cmp"
-	"encoding/base64"
-	"errors"
-	"strconv"
-	"strings"
-
 	"cloud.google.com/go/spanner"
 	sppb "cloud.google.com/go/spanner/apiv1/spannerpb"
 	"github.com/apstndb/spantype"
 	"github.com/apstndb/spanvalue"
+	"github.com/apstndb/spanvalue/protofmt"
 	"google.golang.org/protobuf/encoding/prototext"
-	"google.golang.org/protobuf/proto"
-	"google.golang.org/protobuf/reflect/protodesc"
-	"google.golang.org/protobuf/reflect/protoreflect"
-	"google.golang.org/protobuf/reflect/protoregistry"
 	"google.golang.org/protobuf/types/descriptorpb"
-	"google.golang.org/protobuf/types/dynamicpb"
 )
 
 func DecodeRow(row *spanner.Row) ([]string, error) {
@@ -45,7 +35,7 @@ func DecodeRow(row *spanner.Row) ([]string, error) {
 // If fds is nil, it returns a config without custom proto/enum support.
 // The multiline parameter controls the formatting of protobuf messages.
 func FormatConfigWithProto(fds *descriptorpb.FileDescriptorSet, multiline bool) (*spanvalue.FormatConfig, error) {
-	types, err := dynamicTypesByFDS(fds)
+	resolver, err := protofmt.ProtoEnumResolverFromFileDescriptorSet(fds)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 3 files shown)


internal/mycli/decoder/decoder.go
```diff
@@ -45,95 +35,25 @@ func DecodeRow(row *spanner.Row) ([]string, error) {
 // If fds is nil, it returns a config without custom proto/enum support.
 // The multiline parameter controls the formatting of protobuf messages.
 func FormatConfigWithProto(fds *descriptorpb.FileDescriptorSet, multiline bool) (*spanvalue.FormatConfig, error) {
-	types, err := dynamicTypesByFDS(fds)
+	resolver, err := protofmt.ProtoEnumResolverFromFileDescriptorSet(fds)
```
apstndb, line 40: **Gemini's CRITICAL here is a false positive — no fix needed.** `protofmt.ProtoEnumResolverFromFileDescriptorSet` checks `fds == nil` and returns `dynamicpb.NewTypes(nil), nil` *before* it ever reaches `protodesc.NewFiles`, so the nil-dereference panic the suggestion describes cannot happen. Its godoc confirms: "A nil fds returns an empty resolver with nil error."

With a nil `fds`, the empty resolver makes both `FormatProtoTextValue` and `FormatEnumNameValue` fall through (no types resolved), i.e. "a config without custom proto/enum support" — exactly what the doc comment above promises. Adding the suggested `if fds == nil` early return would only duplicate the upstream guard.

(This preserves the old `dynamicTypesByFDS(nil)` behavior, which used the same `nil → dynamicpb.NewTypes(nil)` shortcut.)

internal/mycli/decoder/decoder.go
```diff
@@ -45,95 +35,25 @@ func DecodeRow(row *spanner.Row) ([]string, error) {
...
 	}
 
 	fc := spanvalue.SpannerCLICompatibleFormatConfig().Clone()
 	fc.FormatComplexPlugins = append(
 		[]spanvalue.FormatComplexFunc{
-			formatProto(types, multiline),
-			formatEnum(types),
+			protofmt.FormatProtoTextValue(protofmt.ProtoTextValueOptions{
+				Resolver: resolver,
+				Marshal:  prototext.MarshalOptions{Multiline: multiline},
```
apstndb, line 50: **Minor parity note (informational).** The old `formatProto` finished with `strings.TrimSpace(...)`, whereas upstream `FormatProtoTextValue` ends with `strings.TrimSuffix(out, "\n")`. For multiline output these are equivalent (single trailing newline). They differ only if non-multiline `prototext` output carries leading/trailing spaces, which `TrimSpace` stripped but `TrimSuffix("\n")` leaves. This is an upstream choice (and arguably safer, since it won't eat meaningful spaces), so nothing to change here — just flagging the observable difference.

apstndb (commented):
Reviewed the branch diff. This is a clean, well-scoped refactor (3 files, +10/−90) that replaces the local proto/enum formatting with `spanvalue/protofmt`, and it holds up — CI is green and behavior parity checks out.

Two things worth recording:

- **Gemini's CRITICAL nil-panic finding is a false positive.** `protofmt.ProtoEnumResolverFromFileDescriptorSet` guards `fds == nil` *before* `protodesc.NewFiles` (returns `dynamicpb.NewTypes(nil), nil`), and its godoc states "A nil fds returns an empty resolver with nil error." So the panic it describes cannot occur, and this function's doc comment ("If fds is nil, …") stays accurate. The suggested early return would only duplicate an upstream guarantee. (Inline detail below.)

- **This refactor also removes a latent nil-panic.** The old `formatEnum` did `cmp.Or(string(enumType.Descriptor().Values().ByNumber(n).Name()), raw)`; for an unknown/out-of-range enum number `ByNumber(n)` returns nil and `.Name()` panics (`cmp.Or` evaluates both args eagerly). The new `protofmt.FormatEnumNameValue` guards `valueDesc == nil` and the numeric range, returning the raw numeric string — which is what the old `cmp.Or` fallback was trying to do. Net safety improvement.

One minor proto-text trimming parity note inline. No blocking issues.

internal/mycli/decoder/decoder.go
apstndb: Addressed in 7246183 with an inline note documenting the upstream nil descriptor behavior, so future reviews should not need to infer it from protofmt internals.

internal/mycli/decoder/decoder.go
```diff
@@ -45,95 +35,25 @@ func DecodeRow(row *spanner.Row) ([]string, error) {
 // If fds is nil, it returns a config without custom proto/enum support.
 // The multiline parameter controls the formatting of protobuf messages.
 func FormatConfigWithProto(fds *descriptorpb.FileDescriptorSet, multiline bool) (*spanvalue.FormatConfig, error) {
-	types, err := dynamicTypesByFDS(fds)
+	resolver, err := protofmt.ProtoEnumResolverFromFileDescriptorSet(fds)
 	if err != nil {
 		return nil, err
 	}
```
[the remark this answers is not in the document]
apstndb: Addressed in 7246183: added an inline note that protofmt preserves the old nil descriptor behavior by returning an empty resolver, so PROTO/ENUM plugins fall through instead of panicking.

PR #656, opened 2026-06-05
Use rendered output for dump fallback

### LINKED ISSUE #649: Remove DUMP buffered fallback from Result.Rows output
## Category

Non-breaking cleanup.

## Background

PR #648 makes SQL export a non-table format that always streams. It also changed the DUMP fallback path to reuse the streaming SQL writer by capturing output into a temporary `StreamManager`.

However, the fallback still returns preformatted text through `Result.Rows` with `IsDirectOutput: true` when there is no real output stream.

## Problem

This leaves a legacy direct-output branch in `printTableData` and keeps DUMP as a special `Result.Rows` producer even though the actual SQL formatting is now stream-based.

## Suggested PR scope

- Replace the DUMP fallback's `Result.Rows`/`IsDirectOutput` return path with a narrower mechanism such as `RenderedOutput`.
- Remove or significantly narrow `IsDirectOutput` if no other production caller needs it.
- Preserve current DUMP output and result-line behavior.
- Preserve the separation between SQL dump output and stderr/warnings.
- Keep PostgreSQL DUMP support out of scope.

## Acceptance criteria

- `DUMP DATABASE`, `DUMP SCHEMA`, and `DUMP TABLES` tests still pass.
- No warnings/errors can be written into generated SQL dump output.
- `printTableData` no longer needs a generic direct-output branch for DUMP fallback, or that branch is documented as intentionally retained for a remaining caller.
- `make check` passes.

## Related

- #648
- #647

apstndb:
## Summary

- route buffered DUMP fallback output through `RenderedOutput` instead of `Result.Rows`
- remove the now-unused `IsDirectOutput` compatibility path from `Result` and `printTableData`
- update DUMP integration tests to assert the pre-rendered output contract

## Why

This keeps DUMP's current output unchanged while avoiding the legacy path that split spanvalue-generated SQL dump text into formatted rows only to print it directly again.

Fixes #649.

## Verification

- `go test ./internal/mycli -run 'TestDump'`
- `make check`

### DIFF
--- a/internal/mycli/statements_dump.go
+++ b/internal/mycli/statements_dump.go
@@ -249,12 +249,11 @@ func writeCapturedDumpOutput(out io.Writer, output string) error {
 	if output == "" {
 		return nil
 	}
-	for _, line := range strings.Split(strings.TrimRight(output, "\n"), "\n") {
-		if _, err := fmt.Fprintln(out, line); err != nil {
-			return err
-		}
+	if _, err := io.WriteString(out, strings.TrimRight(output, "\n")); err != nil {
+		return err
 	}
-	return nil
+	_, err := io.WriteString(out, "\n")
+	return err
 }
 
 func executeDumpTableIntoBuffer(ctx context.Context, session *Session, txn *spanner.ReadOnlyTransaction, selectQuery, table string) (*Result, string, error) {
--- a/internal/mycli/statements_dump.go
+++ b/internal/mycli/statements_dump.go
@@ -162,7 +162,8 @@ func getWritableColumnsWithTxn(ctx context.Context, txn *spanner.ReadOnlyTransac
 // executeDumpBuffered performs dump operation with buffering.
 // All output is collected in memory before being returned.
 func executeDumpBuffered(ctx context.Context, session *Session, mode dumpMode, specificTables []string) (*Result, error) {
-	result := &Result{AffectedRows: 0, IsDirectOutput: true}
+	var out bytes.Buffer
+	var affectedRows int
 
 	// Export DDL first if requested (DDL doesn't need transaction consistency)
 	if mode.shouldExportDDL() {
@@ -170,7 +171,9 @@ func executeDumpBuffered(ctx context.Context, session *Session, mode dumpMode, s
 		if err != nil {
 			return nil, fmt.Errorf("export DDL: %w", err)
 		}
-		result.Rows = append(result.Rows, ddlResult.Rows...)
+		if err := writeResultRows(&out, ddlResult.Rows); err != nil {
+			return nil, fmt.Errorf("write DDL: %w", err)
+		}
 	}
 
 	// Execute all INFORMATION_SCHEMA queries and data export within a single transaction for consistency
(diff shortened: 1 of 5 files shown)


internal/mycli/statements_dump.go
```diff
@@ -240,6 +245,18 @@ func writeResultRows(out io.Writer, rows []Row) error {
 	return nil
 }
 
+func writeCapturedDumpOutput(out io.Writer, output string) error {
+	if output == "" {
+		return nil
+	}
+	for _, line := range strings.Split(strings.TrimRight(output, "\n"), "\n") {
```
apstndb: On the Gemini suggestion to replace this with `io.WriteString(out, output)` — I'd **not apply it verbatim**, because it isn't behavior-preserving. `writeCapturedDumpOutput(out, s)` is exactly `io.WriteString(out, strings.TrimRight(s, "\n") + "\n")`: it collapses trailing newlines to a single one and guarantees a final newline (internal blank lines are preserved). Plain `io.WriteString(out, output)` writes verbatim, so if `dumpOutput` ever lacks a trailing `\n` or carries a trailing blank line, the DUMP output changes — which this PR explicitly sets out to avoid ("keeps DUMP's current output unchanged").

Gemini's underlying point (avoid the per-line `strings.Split` slice) is reasonable, but the impact is marginal here: `executeDumpBuffered` only runs when the output stream is nil/`io.Discard` (real dumps take `executeDumpStreaming`), and the `bytes.Buffer` already holds the whole dump. `strings.Split` also doesn't copy the string data — it allocates one `[]string` header per line, nothing more.

If you do want to drop the split while keeping behavior identical:

```go
func writeCapturedDumpOutput(out io.Writer, output string) error {
	if output == "" {
		return nil
	}
	if _, err := io.WriteString(out, strings.TrimRight(output, "\n")); err != nil {
		return err
	}
	_, err := io.WriteString(out, "\n")
	return err
}
```

`strings.TrimRight` returns a sub-slice (no copy), so this preserves the single-trailing-newline normalization without the per-line slice. Otherwise the current code is correct as-is.

apstndb (commented):
Reviewed the diff. Clean, behavior-preserving refactor (net −21) that drops the `IsDirectOutput` path in favor of the merged `RenderedOutput` mechanism. Parity checks out:

- The DDL path now goes through `writeResultRows`, which is byte-for-byte the same as the removed `IsDirectOutput` printing (`fmt.Fprintln(out, row[0].RawText())` per non-empty row).
- Data lines, the `-- Data for table` / `-- Skipping table` comments, the trailing blank line on non-empty tables, and `AffectedRows` accumulation all map 1:1 to the old appended-`Rows` behavior.
- `IsDirectOutput` is fully removed — no remaining references — and CI is green.

One note on the open Gemini thread, inline.

internal/mycli/statements_dump.go
apstndb: Addressed in f91e940: writeCapturedDumpOutput now uses io.WriteString after TrimRight to avoid allocating the per-line slice while preserving the existing single-trailing-newline normalization. Added unit coverage for empty, missing, extra, internal blank, and newline-only cases.

PR #657, opened 2026-06-10
Adopt spanenc v0.3.0 for client-side virtual result sets

apstndb:
## Summary

- add `github.com/apstndb/spanenc` v0.3.0 and bump `spanvalue` to v0.7.1 (spanenc requirement)
- introduce `spanenc_bridge.go` to build buffered `Result` values from `spanenc.RowEncoder` output via `spanvalue.FormatRowColumns`
- route `SHOW VARIABLES` and `HELP VARIABLES` through the bridge instead of hand-built `toRow` rows
- add unit tests covering spanenc API smoke checks and bridge row shaping

## Why

Client-side virtual result sets (SHOW/HELP) previously built plain string rows manually. spanenc provides compile-once `RowEncoder` codecs that mirror Cloud Spanner client encoding semantics and compose with the spanvalue formatting stack, keeping display behavior aligned with server query results.

## Notes

- FLOAT64/FLOAT32 display expectations change with spanvalue v0.7.1 (`1.000000` → `1`); integration tests updated accordingly.
- Table headers use `RowEncoder.Columns()` (name-only) to preserve existing `simpleTableHeader` rendering.

## Test plan

- [x] `make check`
- [x] `go test ./internal/mycli -run 'Spanenc|ResultFromRowEncoder'`

Made with [Cursor](https://cursor.com)


internal/mycli/spanenc_bridge.go
```diff
@@ -0,0 +1,107 @@
...
+	return decoder.FormatConfigWithProto(
+		session.systemVariables.Internal.ProtoDescriptor,
+		session.systemVariables.Display.MultilineProtoText,
+	)
+}
+
+// resultFromRowEncoder builds a buffered Result from one or more rows of struct
+// type T using a compiled spanenc.RowEncoder. Column names come from the encoder;
+// cell text uses spanvalue.FormatRowColumns on RowEncoder.Values output.
+func resultFromRowEncoder[T any](enc *spanenc.RowEncoder[T], items []T, fc *spanvalue.FormatConfig, typeStyles map[sppb.TypeCode]string, nullStyle string) (*Result, error) {
```
apstndb, line 68: Honest note from the adoption guide's own author: `nameValueRow` / `helpVariableRow` are all-string rows, i.e. the documented "poor fit" end of the guide (the GCV round trip adds work with no display change — the tests confirm output equals the old `toRow` rows). What justifies it here is the scaffolding value: a single styled-cell path shared with server results, and this bridge being the landing zone for genuinely typed call sites later (PROTO-typed virtual columns, MUTATE preview via `MutationColumnsAndValues`). Suggest saying that in this file's header comment so a future cleanup pass doesn't read the adoption guide and "simplify" these statements back to `toRow` — or, if no typed call site is planned, keeping SHOW/HELP on `toRow` and reserving the bridge is equally defensible. Maintainer's call.

internal/mycli/spanenc_bridge.go
```diff
@@ -0,0 +1,107 @@
...
+
+	result := make(Row, len(texts))
+	for i, text := range texts {
+		result[i] = gcvTextToCell(text, gcvs[i], typeStyles, nullStyle)
+	}
+	return result, nil
+}
+
+func gcvTextToCell(text string, gcv spanner.GenericColumnValue, typeStyles map[sppb.TypeCode]string, nullStyle string) format.Cell {
+	if _, isNull := gcv.Value.GetKind().(*structpb.Value_NullValue); isNull {
```
apstndb, line 111: Minor robustness: in gcvctor semantics a GCV is also NULL when `Value` itself is nil, not only on a protobuf `NullValue` kind. spanenc-produced GCVs never carry a nil `Value`, so this is safe today, but `gcvTextToCell` looks reusable for foreign GCVs (e.g. rows read back from the server). Consider `gcv.Value == nil ||` in the condition, or a comment scoping the helper to spanenc-produced values.

internal/mycli/statements_system_variable.go
```diff
@@ -64,14 +64,29 @@ func (s *ShowVariablesStatement) Execute(ctx context.Context, session *Session)
...
-	return &Result{
-		TableHeader:   toTableHeader("name", "value"),
-		Rows:          rows,
-		KeepVariables: true,
-	}, nil
+	fc, err := clientSideFormatConfig(session)
+	if err != nil {
+		return nil, err
+	}
+	result, err := resultFromRowEncoder(
```
apstndb, line 78: Behavior delta to verify: the previous `ShowVariablesStatement` Result did not set `AffectedRows` (it set only TableHeader/Rows/KeepVariables), while `resultFromRowEncoder` always sets `AffectedRows = len(rows)`. If any output path renders a row-count footer from `AffectedRows`, SHOW VARIABLES output changes. HELP VARIABLES already set it before, so only this call site is affected. If it turns out to matter, a `withAffectedRows bool` knob is uglier than just clearing it here (`result.AffectedRows = 0`).

internal/mycli/spanenc_bridge_test.go
```diff
@@ -0,0 +1,172 @@
...
+	"testing"
+
+	sppb "cloud.google.com/go/spanner/apiv1/spannerpb"
+	"github.com/apstndb/spanenc"
+	"github.com/apstndb/spanner-mycli/internal/mycli/decoder"
+	"github.com/apstndb/spanvalue"
+	"github.com/google/go-cmp/cmp"
+)
+
+func TestSpanencAPIOverview(t *testing.T) {
```
apstndb: Non-blocking: `TestSpanencAPIOverview` pins spanenc's own API behavior (ParamsMap read-only inclusion, MutationMap exclusion, slice helpers) inside the consumer repo, duplicating spanenc's test suite. While spanenc is experimental that's churn on every upgrade — suggest trimming it to the surface the bridge actually uses (`RowEncoder` + `FormatRowColumns`), or renaming/commenting it as intentional adoption documentation if the broad pin is the point.

apstndb (commented):
Reviewed as the spanenc author. The integration uses the library exactly as intended: package-level compile-once `RowEncoder`s, `Columns()` for name-only headers (per the adoption guide, this keeps `simpleTableHeader` and existing goldens stable), and `spanvalue.FormatRowColumns` instead of a per-app GCV→string bridge — which is precisely the fix for the PoC's `gcvToCell` duplication feedback (spanenc#1). The FLOAT64/FLOAT32 golden updates (`1.000000` → `1`) are the documented spanvalue v0.7.1 formatting coupling from the [v0.3.0 release notes](https://github.com/apstndb/spanenc/releases/tag/v0.3.0), unrelated to spanenc itself.

Four inline notes; nothing blocking from the spanenc side.

internal/mycli/spanenc_bridge.go
apstndb: Correction to my own suggestion: no hand-rolled check is needed at all — `spanvalue.IsNull(gcv)` is already exported (root package) and implements exactly these semantics (nil `Value` or protobuf `NullValue`). Use that instead of inspecting the protobuf kind; I've also added the pointer to spanenc's adoption guide so the next adopter finds it (apstndb/spanenc@main).

internal/mycli/spanenc_bridge.go
```diff
@@ -0,0 +1,120 @@
...
+	Name  string `spanner:"name"`
+	Value string `spanner:"value"`
+}
+
+var (
+	helpVariablesRowEncoder = mustNewRowEncoder[helpVariableRow]()
+	nameValueRowEncoder     = mustNewRowEncoder[nameValueRow]()
+)
+
+func mustNewRowEncoder[T any]() *spanenc.RowEncoder[T] {
```
apstndb, line 46: Follow-up from the spanenc side: this local helper prompted adding `spanenc.MustNewRowEncoder[T](opts...)` upstream (on main, will ship in the next release) — same rationale as gcvctor's Must* family for compile-time-known types. Once you bump spanenc, this wrapper can be deleted in favor of `var helpVariablesRowEncoder = spanenc.MustNewRowEncoder[helpVariableRow]()`.

apstndb:
Superseded by #661, which replaces the GCV-level cell formatting with real *spanner.Row encoding through the released spanenc v0.3.1 / spanvalue v0.7.4 APIs (RowEncoder.Rows, writer.WriteRowSeq) and adds CSV/JSONL streaming for client-side virtual result sets. Closing for a clean review history.

PR #661, opened 2026-06-11
Render client-side virtual result sets through the server result pipeline

apstndb:
## Summary

Supersedes #657 with a cleaner design and the released upstream APIs (fresh PR for a clean review history).

Client-side (virtual) result sets — SHOW VARIABLES / HELP VARIABLES — are now defined as Go structs and rendered by the exact same code paths as server query results:

- **Row definition**: `spanenc.RowEncoder` (v0.3.1, package-level via `MustNewRowEncoder`) compiles the row shape once; `RowEncoder.Rows` encodes structs into real `*spanner.Row` values. `spanner.NewRow` deep-clones `GenericColumnValue` inputs without re-encoding, so rows carry exactly the encoder output including typed NULLs.
- **CSV / JSONL**: `streamStructRows` streams rows through the same spanvalue writers as the query streaming path via `writer.WriteRowSeq` (spanvalue v0.7.4), sharing writer construction (`newSpanvalueRowIteratorWriterFor`, extracted from `newSpanvalueRowIteratorWriter`). SHOW/HELP VARIABLES output in these modes no longer buffers.
- **Other formats**: `resultFromStructRows` buffers cells through `spannerRowToRow` / `withRawJSONMarker` — the row transforms the query path uses — instead of a parallel cell-formatting implementation. `clientSideFormatContext` mirrors `prepareFormatConfig` including the JSONValues branch, so typed virtual columns render as native JSON in JSONL.
- **Headers**: table headers carry `RowEncoder.ResultSetMetadata` row types (`typesTableHeader`), so verbose mode shows column types for virtual result sets like it does for queries.

## Why

#657 routed SHOW/HELP VARIABLES through spanenc but duplicated the cell-formatting layer (`gcvTextToCell` vs `spannerRowToRow`) and only the display branch of the FormatConfig derivation. Encoding to real `*spanner.Row` values makes the rendering identical to server results **by construction** rather than by parallel implementation. The `RowEncoder.Row`/`Rows` and `writer.RunRowSeq`/`WriteRowSeq` APIs were upstreamed from this work (apstndb/spanenc#2, apstndb/spanvalue#238) and released in spanenc v0.3.1 / spanvalue v0.7.4.

## Behavior changes

- SHOW/HELP VARIABLES in `--verbose` table mode now show column types in headers (consistent with query results); integration test expectation updated.
- SHOW/HELP VARIABLES in CSV/JSONL stream through spanvalue writers (byte-identical output for the current all-STRING columns; future typed columns get native JSON values in JSONL).
- FLOAT64/FLOAT32 display changes with spanvalue v0.7.x (`1.000000` → `1`), inherited from the dependency bump; integration tests updated.
- SQL export modes (`SQL_INSERT` etc.) deliberately keep the buffered table-format fallback: client-side results have no source table.
- SHOW VARIABLES now sets `AffectedRows` (previously unset), aligning with HELP VARIABLES: the interactive result line changes from `Empty set` to `N rows in set`.

## Test plan

- [x] `make check` (test + lint + fmt-check), including emulator integration tests
- [x] New unit tests: streaming (CSV/JSONL output bytes, TABLE fallback), typed header metadata, JSONL native values (`RawJSONCell`), display-mode typed NULL handling
- [x] Manual: `SHOW VARIABLES` against embedded emulator in JSONL / CSV / TABLE / verbose TABLE


internal/mycli/spanenc_bridge_test.go
```diff
@@ -0,0 +1,318 @@
...
+	sppb "cloud.google.com/go/spanner/apiv1/spannerpb"
+	"github.com/apstndb/spanenc"
+	"github.com/apstndb/spanner-mycli/enums"
+	"github.com/apstndb/spanner-mycli/internal/mycli/format"
+	"github.com/apstndb/spanner-mycli/internal/mycli/streamio"
+	"github.com/apstndb/spanvalue"
+	"github.com/google/go-cmp/cmp"
+)
+
+func TestSpanencAPIOverview(t *testing.T) {
```
apstndb: This test exercises spanenc APIs this PR does not use anywhere (`ValueOf`, `StructColumns`, `RowTypeFor`, `ResultSetMetadataFor`, `ParamsMap`, `MutationMap`, `ValuesFromSlice`, `ArrayValueFromSlice`) — only `RowEncoder` / `Rows` / `ResultSetMetadata` back the bridge. Each assertion duplicates coverage that already lives in spanenc's own test suite and pkg.go.dev examples, so in mycli it pins nothing app-specific while adding drift cost on every spanenc bump. Suggest trimming it to the surface the bridge consumes (or dropping it — the bridge tests below already exercise that surface end-to-end). If the intent is an early-warning canary against dependency bumps, a one-line comment saying so would stop future cleanups from deleting it for the same reason I'm raising it.

internal/mycli/statements_system_variable.go
```diff
@@ -64,14 +64,19 @@ func (s *ShowVariablesStatement) Execute(ctx context.Context, session *Session)
...
+	slices.SortFunc(items, func(lhs, rhs nameValueRow) int {
+		return cmp.Compare(lhs.Name, rhs.Name)
+	})
 
-	return &Result{
-		TableHeader:   toTableHeader("name", "value"),
-		Rows:          rows,
-		KeepVariables: true,
-	}, nil
+	result, err := executeStructRows(nameValueRowEncoder, items, session.systemVariables)
```
apstndb, line 74: Behavior question not covered in the PR body: the old `ShowVariablesStatement` returned a `Result` without setting `AffectedRows`, while `executeStructRows` now sets it (`len(rows)` buffered / `RowsRead` streamed). HELP VARIABLES already set it before, so this aligns the two — likely an improvement — but if anything renders a row-count footer from `AffectedRows`, SHOW VARIABLES output changes. If intentional, consider adding it to the Behavior changes list; the integration test deliberately skips `AffectedRows` ("Rows and AffectedRows are dynamic"), so nothing pins it either way.

apstndb (commented):
Review from the spanenc/spanvalue side (I maintain both and reviewed apstndb/spanenc#2 / apstndb/spanvalue#238 that this PR consumes).

The library usage is exactly as designed, and the claims the PR makes on behalf of the dependencies are accurate: `spanner.NewRow` deep-clones `GenericColumnValue` inputs without re-encoding (verified against client v1.91.0 during the spanenc#2 review), `RowEncoder.Rows` laziness and its yield-once-then-stop error contract are pinned by spanenc's own tests, and `WriteRowSeq`'s abort-on-yielded-error / header-only-empty behavior matches what `streamStructRows` relies on. Sharing `newSpanvalueRowIteratorWriterFor` between the query and virtual paths is the right extraction — writer construction stays in the app, per the spanvalue adoption boundary.

One upstream follow-up this PR surveyed into existence: apstndb/spanenc#3 adds `RowEncoder.MustResultSetMetadata()` so package-level encoders can hoist metadata next to `MustNewRowEncoder` (row-type resolution is deterministic per T + mask). Optional here — with the current generic `executeStructRows[T]` helper shape, keeping the fallible accessor per call is also reasonable — so adopt it on a later bump only if you restructure toward package-level metadata.

Two inline comments below; neither blocks.

internal/mycli/spanenc_bridge_test.go
apstndb: Dropped TestSpanencAPIOverview entirely. The bridge tests below cover the surface this PR consumes (RowEncoder.Rows / ResultSetMetadata) end-to-end, and the broader API coverage belongs to spanenc's own suite and examples as you noted. No canary intent worth preserving — the latest-deps style drift detection is better done on the spanenc side.

Fixed in commit a4792fb.

internal/mycli/statements_system_variable.go
apstndb: Intentional — it aligns SHOW VARIABLES with HELP VARIABLES, and it actually fixes a footer bug: with AffectedRows unset, the interactive result line printed "Empty set" under the rendered rows (resultLine treats TableHeader != nil with AffectedRows == 0 as an empty result set). Now it prints "N rows in set". Added to the Behavior changes list in the PR body. No code change needed.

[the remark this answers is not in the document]
apstndb: No change needed, for two reasons:

1. The suggested guard would not prevent the panic it describes: Execute dereferences session.systemVariables at function entry (ListVariables, plus the COMMIT_RESPONSE / CLI_DIRECT_READ handling) well before this line.
2. SHOW VARIABLES reads per-session state, so it is never executed without a session — this is unchanged from the pre-PR code, which made the same dereferences. The nil-session tolerance in HelpVariablesStatement is deliberate and specific to it: HELP VARIABLES is also invoked without a session (registry-only output via newSystemVariablesWithDefaults), while SHOW VARIABLES has no meaning without one.

PR #689, opened 2026-07-01
Add BIGQUERY client-side statement

apstndb:
## Summary
- Add `BIGQUERY <sql>` client-side statement for executing BigQuery SQL from spanner-mycli (works in batch mode and detached mode)
- Add `CLI_BIGQUERY_PROJECT`, `CLI_BIGQUERY_LOCATION`, and `CLI_BIGQUERY_MAX_BYTES_BILLED` system variables
- Use BigQuery-specific auth options (credential/ADCPlus/impersonation shared with Spanner, but no emulator endpoint or `WithoutAuthentication`)
- Format BigQuery values into the existing result table pipeline, including headers on 0-row results

## Test plan
- [x] `make check`
- [ ] Manual: `BIGQUERY SELECT 1` against a real GCP project
- [ ] Manual: `BIGQUERY SELECT 1 WHERE FALSE` confirms header + `Empty set`
- [ ] Manual: verify `CLI_BIGQUERY_MAX_BYTES_BILLED` is applied

Made with [Cursor](https://cursor.com)


internal/mycli/session.go
```diff
@@ -1020,10 +1076,21 @@ func createSession(ctx context.Context, credential []byte, sysVars *systemVariab
 		return nil, err
 	}
 
+	bqOpts, err := createBigQueryClientOptions(ctx, credential, sysVars)
```
apstndb: This eagerly builds BigQuery auth options for every session. With ADCPlus enabled and no ADC, createSession can fail for ordinary Spanner/emulator sessions before any BIGQUERY statement is used. Please build BigQuery auth lazily on first BigQuery use or otherwise keep this path out of normal session creation.

internal/mycli/session.go
```diff
@@ -969,24 +1004,45 @@ func createClientOptions(ctx context.Context, credential []byte, sysVars *system
...
+
+func bigQueryProject(sysVars *systemVariables) string {
+	if p := sysVars.Feature.BigQueryProject; p != "" {
+		return p
+	}
+	return sysVars.Connection.Project
+}
+
+func (s *Session) bigQueryClient(ctx context.Context) (*bigquery.Client, error) {
+	if s.bqClient != nil {
```
apstndb: Caching a single client makes CLI_BIGQUERY_PROJECT effectively immutable after the first BIGQUERY query. SET CLI_BIGQUERY_PROJECT later will continue using the old client/project. Either cache by project or document and enforce session-init-only behavior.

internal/mycli/statements_bigquery.go
```diff
@@ -0,0 +1,165 @@
...
+		return val.String()
+	case civil.Time:
+		return val.String()
+	case civil.DateTime:
+		return val.String()
+	case *big.Rat:
+		if val == nil {
+			return "NULL"
+		}
+		return val.FloatString(9)
```
apstndb, line 135: This loses precision for BIGNUMERIC. BigQuery returns both NUMERIC and BIGNUMERIC as *big.Rat, but FloatString(9) caps output at NUMERIC scale. Please format using the schema field type, e.g. NumericString vs BigNumericString, or otherwise preserve the exact rational value.

internal/mycli/statements_bigquery.go
```diff
@@ -0,0 +1,165 @@
+package mycli
```
apstndb, line 15: New Go source files in this repository should include the license header before package. This file and the new test file are missing it.

apstndb (commented):
Blocking findings: BigQuery auth is eager for every session, BIGNUMERIC precision is lost, project changes are cached after first use, and new files need license headers.

internal/mycli/session.go
apstndb: Fixed: BigQuery client options are now built lazily on the first BIGQUERY statement in Session.bigQueryClient (internal/mycli/session.go); createSession only retains the raw credential and never resolves BigQuery auth. Ordinary Spanner/emulator sessions no longer touch BigQuery credentials; TestRun/embedded_emulator_success passes credential-less.

Fixed in commit e57e1b5.

internal/mycli/session.go
apstndb: Fixed: the BigQuery client is now cached keyed by (CLI_BIGQUERY_PROJECT, CLI_BIGQUERY_LOCATION) via bigQueryClientKey (internal/mycli/session.go). On each BIGQUERY statement, if the current variable values differ from the cached client's key, the old client is closed and a new one built. SET of these variables stays runtime-allowed. The lock-free check-then-act is safe because BIGQUERY runs on the single-threaded statement path (same rationale as cqlSession, commented at the call site).

Fixed in commit e57e1b5.

internal/mycli/statements_bigquery.go
apstndb: Fixed: formatBigQueryValue now takes the schema field type and formats *big.Rat by it (internal/mycli/statements_bigquery.go): FloatString(38) for BIGNUMERIC, FloatString(9) for NUMERIC. The Execute loop passes it.Schema[i].Type. Added a BIGNUMERIC test case asserting full scale-38 precision.

Fixed in commit e57e1b5.

internal/mycli/statements_bigquery.go
apstndb: Fixed: added the Apache 2.0 license header ("Copyright 2026 apstndb") before package to both internal/mycli/statements_bigquery.go and internal/mycli/statements_bigquery_test.go.

Fixed in commit e57e1b5.

Issue #647, opened 2026-06-04
Migrate buffered export output toward spanvalue writer


apstndb:
## Background

PR #645 moved streaming CSV, JSONL, and SQL export paths to spanvalue RowIterator writers. Buffered execution still collects already formatted `Result.Rows` (`[]Row` / `[]Cell`) and renders through spanner-mycli formatters.

## Problem

spanvalue writer APIs can write rows from raw `spanner.GenericColumnValue` values, but the current buffered path converts Spanner rows to formatted cells before output. At that point schema and raw value details needed by spanvalue writer are no longer directly available.

## Scope

- Investigate a buffered representation that preserves metadata plus raw row values, such as `[][]spanner.GenericColumnValue`.
- Route buffered CSV, JSONL, and SQL export output through spanvalue writer APIs where practical.
- Preserve `CLI_STREAMING=false`, SQL table-name handling, SQL batching, unnamed column behavior, and result/appendix output.
- Keep TABLE, VERTICAL, HTML, and XML out of scope unless a clear writer boundary emerges.
- Treat raw TAB output separately from spanvalue delimited TSV, because spanvalue delimited output uses CSV-style quoting rules.

## Notes

TABLE-style output still depends on spanner-mycli-specific width calculation, ANSI styling, verbose headers, result lines, and appendices. Those should not be forced into spanvalue writer as part of this follow-up.

apstndb:
Follow-up from #648 output-path cleanup triage.

This issue remains relevant and should be treated as a non-breaking output-formatting follow-up. It should stay separate from the breaking changes around existing mode/variable names.

apstndb:
Premise update: after #648, non-table formats *always* stream (`createStreamingProcessor`, execute_sql.go) — `executeWithBuffering` only runs for table format or a nil writer — so the buffered-export representation question mostly evaporated. For the residual buffered-replay cases (client-side tables, partitioned query), do not invent a `[][]GenericColumnValue` representation: buffer `[]*spanner.Row` (lossless — `spanner.NewRow` deep-clones GCVs including typed NULLs) and replay through spanvalue v0.7.4's `writer.WriteRowSeq`/`RunRowSeq` with the existing `newSpanvalueRowIteratorWriterFor`, exactly the pattern #661 merged for virtual result sets (`streamStructRows`). Metadata travels in `ResultSetMetadata`, and writer construction stays single-sourced. Suggest rescoping this issue to that residual, or closing in favor of #651.

[...]

PR #711, opened 2026-07-05
Add TSV output format with lossless escaping

### LINKED ISSUE #644: Define true TSV output semantics
## Background

The current `TAB` formatter writes rows by joining already-formatted cell text with literal tab characters:

```go
strings.Join(Texts(row), "\t")
```

This is simple and useful for scripting, but it is not a robust TSV export format when values can contain tabs, newlines, quotes, or other field-boundary-sensitive characters.

While considering typed-row export formatting with `spanvalue/writer` in #643, the `TAB` path needs an explicit compatibility decision. `spanvalue.NewDelimitedWriter(out, '\t', ...)` uses CSV-style delimited escaping with a tab delimiter, which would be more robust but can change output for values that need quoting.

## Proposal

Define and support a "true TSV" export mode or option before moving `TAB` onto spanvalue's delimited writer path.

Open questions:

- Should existing `CLI_FORMAT=TAB` keep the current raw `strings.Join` behavior for backward compatibility?
- Should a new mode such as `TSV` be added for escaped tab-delimited output?
- Should `TAB` itself become the escaped TSV mode as a breaking change?
- Which escaping convention should be documented: CSV-style quoting with tab delimiter, backslash escaping, or another convention?

## Acceptance criteria

- The expected behavior for embedded tabs, newlines, quotes, and empty strings is documented.
- Tests cover fields that contain tab and newline characters.
- The decision is reflected in the typed-row export plan in #643.

## Related

- #643
- #203
- apstndb/spanvalue#98

apstndb:
## Summary

Fixes #644

Adds a new `CLI_FORMAT` value `TSV` alongside the existing `TAB` format. `TAB` is a raw tab-join of formatted cells and silently breaks the row/column structure when values contain tabs or newlines. Rather than changing `TAB` (backward compatibility), this PR defines true TSV semantics as a separate format with conventional lossless escaping (as used by MySQL tab output / PostgreSQL COPY text format).

`--format=tsv` and `SET CLI_FORMAT = 'TSV'` are both supported. TSV streams like other non-table formats and works for both server query results and client-side statements.

**Note:** #647 (typed buffered export) is intentionally out of scope — TSV here operates on the existing display-formatted cell pipeline, same as TAB.

## Semantics

| Input in value | TAB (unchanged) | TSV (new) |
|---|---|---|
| tab (`U+0009`) | emitted raw (breaks columns) | `\t` |
| newline (`U+000A`) | emitted raw (breaks rows) | `\n` |
| carriage return (`U+000D`) | emitted raw | `\r` |
| backslash (`\`) | emitted raw (ambiguous) | `\\` |
| double/single quotes | emitted raw | emitted raw (quotes are not special in TSV) |
| empty string | empty field | empty field |
| NULL | literal `NULL` text | literal `NULL` text (same display convention as TAB/TABLE; no new NULL literal) |

Header names are escaped the same way as values (aliases can contain tabs/newlines). Because backslash is escaped, output is losslessly round-trippable: `\t` in output always means an escaped tab, never a literal backslash-t.

## Implementation

- `enums`: add `DisplayModeTSV`, regenerate enumer code — `SET CLI_FORMAT`, `--format`, and config-file parsing accept `TSV`/`tsv` automatically
- `internal/mycli/format`: add `ModeTSV` and `TSVFormatter` (mirrors `TabFormatter` with an escaping `strings.Replacer`); wired into `NewStreamingFormatter`, which serves both the streaming and buffered output paths
- `CLI_FORMAT` sysvar description and `--format` flag help updated; README help regenerated with `make docs-update`
- Docs: TAB vs TSV semantics documented in `README.md` and `docs/system_variables.md`
- No `--tsv` shorthand flag: the individual format flags exist for spanner-cli compatibility; `--format=tsv` is sufficient

## Tests

- `TestFormatTSV`: table-driven cases for tab, newline, CR (incl. CRLF), backslash (incl. literal `\t` text), quotes-verbatim, empty string, `NULL` literal, header-name escaping, `--skip-column-names`, empty result
- `TestFormatTab`: new regression case asserting TAB's raw pass-through is byte-for-byte unchanged
- `printResult` end-to-end test with `CLI_FORMAT=TSV` (header + escaped rows)
- TSV added to formatter capability, lifecycle, value-format-mode, and ANSI-styling-stripped test matrices; `CLI_FORMAT` set/get tests cover `TSV`/`tsv`
- Verified locally: `go test -short ./...`, `golangci-lint run`, gofumpt/goimports on changed files (the only failures are pre-existing and unrelated: a root-environment permission test and a gofumpt finding in `transaction_manager.go` present on main)

https://claude.ai/code/session_01Xf5L7KM76D3GkTacNDWcg3

---
_Generated by [Claude Code](https://claude.ai/code/session_01Xf5L7KM76D3GkTacNDWcg3)_

### DIFF
--- a/docs/system_variables.md
+++ b/docs/system_variables.md
@@ -19,7 +19,7 @@ TODO
   SELECT * FROM users;  -- Output without column headers
   ```
 - **Notes**:
-  - Affects table format (`CLI_FORMAT='TABLE'`), tab format (`CLI_FORMAT='TAB'`), CSV format (`CLI_FORMAT='CSV'`), HTML format (`CLI_FORMAT='HTML'`), and XML format (`CLI_FORMAT='XML'`)
+  - Affects table format (`CLI_FORMAT='TABLE'`), tab format (`CLI_FORMAT='TAB'`), TSV format (`CLI_FORMAT='TSV'`), CSV format (`CLI_FORMAT='CSV'`), HTML format (`CLI_FORMAT='HTML'`), and XML format (`CLI_FORMAT='XML'`)
   - Headers are always preserved in vertical format (`CLI_FORMAT='VERTICAL'`) as they are integral to the format
   - Can be set via `--skip-column-names` command-line flag
   - Useful for scripting and data processing where only raw data is needed
--- a/docs/system_variables.md
+++ b/docs/system_variables.md
@@ -72,7 +72,8 @@ TODO
   - `TABLE_COMMENT` - Table wrapped in `/* */` comments
   - `TABLE_DETAIL_COMMENT` - Table and execution details wrapped in `/* */` comments (useful for embedding results in SQL code blocks)
   - `VERTICAL` - Vertical format with column:value pairs
-  - `TAB` - Tab-separated values
+  - `TAB` - Tab-separated values (raw; values are joined with tabs as-is, so values containing tabs or newlines break the row/column structure)
+  - `TSV` - Tab-separated values with escaping (tab, newline, carriage return, and backslash inside values are escaped as `\t`, `\n`, `\r`, and `\\`, guaranteeing one row per line and one field per column)
   - `HTML` - HTML table format
   - `XML` - XML format
   - `CSV` - Comma-separated values (RFC 4180 compliant)
@@ -109,6 +110,7 @@ TODO
   - HTML and XML formats are compatible with Google Cloud Spanner CLI
   - All special characters are properly escaped in HTML, XML, and CSV formats for security
   - CSV format follows RFC 4180 standard with automatic escaping of commas, quotes, and newlines
+  - TSV format escapes `\` as `\\`, tab as `\t`, newline as `\n`, and carriage return as `\r` in both header names and values; double quotes are not special and are emitted verbatim; empty strings produce empty fields; NULL is rendered as the literal text `NULL` (same as TAB/TABLE). Use TAB if you need the historical raw (unescaped) behavior
   - JSONL format produces type-aware JSON: INT64/ENUM as numbers, BOOL as booleans, ARRAY as JSON arrays, STRUCT as JSON objects, NULL as null
   - The format affects how query results are displayed, not how they are executed
   - `TABLE_DETAIL_COMMENT` is particularly useful with `CLI_ECHO_INPUT=TRUE` and `CLI_MARKDOWN_CODEBLOCK=TRUE` for documentation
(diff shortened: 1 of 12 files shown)


docs/system_variables.md
```diff
@@ -72,7 +72,8 @@ TODO
   - `TABLE_COMMENT` - Table wrapped in `/* */` comments
   - `TABLE_DETAIL_COMMENT` - Table and execution details wrapped in `/* */` comments (useful for embedding results in SQL code blocks)
   - `VERTICAL` - Vertical format with column:value pairs
-  - `TAB` - Tab-separated values
+  - `TAB` - Tab-separated values (raw; values are joined with tabs as-is, so values containing tabs or newlines break the row/column structure)
+  - `TSV` - Tab-separated values with escaping (tab, newline, carriage return, and backslash inside values are escaped as `\t`, `\n`, `\r`, and `\\`, guaranteeing one row per line and one field per column)
```
apstndb, line 76: Since TSV honors skipHeaders like TAB and CSV, the CLI_SKIP_COLUMN_NAMES documentation earlier in this file also needs TSV in its affected-format list. Otherwise docs say the new formatter is excluded even though it works.

internal/mycli/var_registry.go
```diff
@@ -335,7 +335,7 @@ func (r *VarRegistry) registerAll() {
 	r.Register("AUTOCOMMIT_DML_MODE", AutocommitDMLModeVar(&sv.Transaction.AutocommitDMLMode,
 		"A STRING property indicating the autocommit mode for Data Manipulation Language (DML) statements."))
 	r.Register("CLI_FORMAT", DisplayModeVar(&sv.Display.CLIFormat,
-		"Controls output format for query results. Valid values: TABLE (ASCII table), TABLE_COMMENT (table in comments), TABLE_DETAIL_COMMENT, VERTICAL (column:value pairs), TAB (tab-separated), HTML (HTML table), XML (XML format), CSV (comma-separated values)."))
+		"Controls output format for query results. Valid values: TABLE (ASCII table), TABLE_COMMENT (table in comments), TABLE_DETAIL_COMMENT, VERTICAL (column:value pairs), TAB (tab-separated, raw values), TSV (tab-separated with tab/newline/backslash escaping), HTML (HTML table), XML (XML format), CSV (comma-separated values)."))
```
apstndb: The implementation also escapes carriage return as \r, but this generated help text only mentions tab/newline/backslash. Please include carriage return here so generated docs and help stay accurate.

apstndb (commented):
Small docs/help nits for TSV documentation.

docs/system_variables.md
apstndb: Fixed in 104c3d5: added TSV (`CLI_FORMAT='TSV'`) to the CLI_SKIP_COLUMN_NAMES affected-format list, between TAB and CSV.

---
_Generated by [Claude Code](https://claude.ai/code)_

internal/mycli/var_registry.go
apstndb: Fixed in 104c3d5: the CLI_FORMAT description now reads "TSV (tab-separated with tab/newline/carriage-return/backslash escaping)", matching the implementation. Checked the other escaping-set mentions (README.md, docs/system_variables.md TSV sections) — they already listed carriage return, so only this spot needed the fix. `make docs-update` produced no further diff.

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #652, opened 2026-06-04
Consolidate client-side tabular outputs behind a shared helper


apstndb:
## Category

Non-breaking refactor.

## Background

PR #648 removes the old buffered formatter registry and keeps regular non-table query output streaming. `Result.Rows` remains widely used for small client-side or metadata-style tables such as system variables, params, proto/schema statements, query profile output, and EXPLAIN/DESCRIBE helpers.

These are generally small outputs, so this is not a memory-safety issue.

## Problem

Ad hoc `Result{Rows: ...}` construction keeps `printTableData` as a broad compatibility adapter and makes it harder to reason about which paths are true query streaming paths versus small client-side tables.

## Suggested PR scope

- Introduce or consolidate a helper for small client-side tabular outputs.
- Migrate a coherent set of statement handlers in one PR, for example system variables/params/proto listing, while keeping diff size reviewable.
- Preserve all existing display formats and result-line behavior.
- Do not force TABLE rendering or spanvalue writer ownership into this cleanup.
- Leave large query paths and partitioned query paths to separate issues.

## Acceptance criteria

- The migrated client-side statements no longer hand-roll `Result.Rows` construction at each call site.
- Output snapshots/expectations remain unchanged.
- The helper makes it explicit that these are small, already-materialized client-side result tables.
- `make check` passes.

## Related

- #648

apstndb:
Direction update: the consolidation point this issue asked for now exists — `executeStructRows[T]` (spanenc_bridge.go, #661), which also brings CSV/JSONL streaming and typed verbose headers for free. Introducing a second, weaker string-row helper would create a parallel abstraction that #660 immediately obsoletes. Suggest narrowing this issue to: migrate the fixed-shape `toRow` call sites to `executeStructRows` (that is #660's work list) and document the residual genuinely-dynamic cases (singular SHOW VARIABLE, params, pre-rendered DDL text) that stay on `toRow`/`printTableData` — or simply merging this into #660.

apstndb:
Concrete design input from a full read-through of the output layer (2026-07; symbol references current as of main after #692).

## Why this issue should land before #647

[...]

PR #715, opened 2026-07-05
Route streamed statement output through a per-statement output context

apstndb:
## Summary

Streaming formats (CSV/JSONL/SQL_INSERT*), streamed table output (`CLI_TABLE_STREAMING=TRUE`), and DUMP wrote to the process-global StreamManager writer regardless of the writer passed to `Cli.executeStatement`. Under `--mcp` that writer is the JSON-RPC stdout, so any streamed statement corrupted the MCP protocol stream instead of appearing in the tool result.

- Add `outputContext` (writer + lazily resolved screen width), resolved once per statement in `Cli.executeStatement` and carried by the `Session` for the duration of one statement; re-entrant executions (RUN BATCH) inherit it.
- Route `decideExecutionMode`, the spanvalue writers, client-side virtual result sets, partitioned query streaming, and DUMP through `Session.outputWriter()`, keeping the StreamManager writer as the fallback for direct `Session.ExecuteStatement` callers.
- Replace DUMP's temporary in-place StreamManager swap with a scoped `withOutput`, removing the last mutation of `systemVariables.StreamManager` during execution.
- Unify the streamed/buffered screen width source on `Cli.resolveScreenWidth` (previously streaming used `StreamManager.GetTerminalWidth()` while buffered display used `GetTerminalSizeWithTty`, which could disagree in redirected-output edge cases).

This is the first step of the output-layer consolidation direction discussed in #652 / #647: the output destination becomes an explicit per-statement parameter instead of ambient global state.

## Behavior changes

- Under `--mcp`, streamed statement output now appears in the `execute_statement` tool result instead of corrupting the JSON-RPC stream.
- Streamed table rendering now resolves its width from the same source as buffered rendering.

## Validation

- `make check` (full test suite with emulator, lint, fmt-check) passes locally.
- `go vet -tags integration,skip_slow_test ./...` passes.
- New regression tests: per-statement routing, StreamManager fallback, nested `withOutput` restore, and an MCP handler test asserting streamed rows land in the tool result with zero bytes leaked to the global stream.


internal/mycli/output_context.go
```diff
@@ -0,0 +1,106 @@
...
+
+// ExecuteStatementWithOutput executes a statement like ExecuteStatement,
+// routing streamed output to out. Session-changing statements (USE/DETACH)
+// produce no streamed output and take their normal path.
+func (h *SessionHandler) ExecuteStatementWithOutput(ctx context.Context, stmt Statement, out outputContext) (*Result, error) {
+	switch stmt.(type) {
+	case *UseStatement, *UseDatabaseMetaCommand, *DetachStatement:
+		return h.ExecuteStatement(ctx, stmt)
+	default:
+		return h.Session.ExecuteStatementWithOutput(ctx, stmt, out)
```
apstndb, line 104: This still leaves ShellMetaCommand outside the per-statement output path. The MCP handler can parse and execute \! commands while system commands are enabled by default, and ShellMetaCommand writes stdout to session.systemVariables.StreamManager.GetWriter. Under --mcp that is the JSON-RPC stdout, so \! echo hi can still corrupt the protocol even though other streamed output is routed through out. Please route shell stdout through session.outputWriter or block shell meta commands in MCP.

apstndb (commented):
One remaining MCP output-routing gap: shell meta command stdout still bypasses the per-statement output context.

internal/mycli/output_context.go
apstndb: Thank you for the feedback!

Fixed in commit 8e4db97.

PR #752, opened 2026-07-06
fix: surface operation ID and invalidate schema on canceled DDL wait

apstndb:
## Summary

Improves the cancellation and error behavior of the synchronous DDL wait loop in `internal/mycli/execute_ddl.go`. When a user cancels a running DDL wait (Ctrl+C), the submitted `UpdateDatabaseDdl` long-running operation (LRO) keeps executing server-side. This change gives the user a handle to that operation and keeps the schema cache honest on every error exit.

## Motivation

The DDL polling loop is context-aware, but on cancellation it had two problems:

1. **No handle to the still-running operation.** The loop returned a bare context error, so the user had no way to find the operation that was still running server-side, even though the CLI already provides `SHOW OPERATION <id>` for exactly this purpose.
2. **Stale schema cache.** `IncrementSchemaGeneration()` (which signals that schema-dependent caches should be invalidated) was only called on the success path and the async path — never on cancellation or other error exits. Because the DDL was already submitted and may still complete server-side, cancelling left the session's schema cache stale.

## Current vs. new behavior

Both `DdlStatement` and `BulkDdlStatement` execute through `executeDdlStatements`, so both share this wait loop and both are covered.

Before — on Ctrl+C during the wait:

```
context canceled
```

(no operation handle; schema cache not invalidated)

After — on Ctrl+C during the wait:

```
stopped waiting for DDL to complete; the operation may still be running server-side, check it with: SHOW OPERATION '1234567890': context canceled
```

The user can then run `SHOW OPERATION '1234567890'` (or `SHOW OPERATION '<id>' SYNC`) to attach to and monitor the operation. The operation ID is extracted with the same formatting used by the async DDL result path (`lo.LastOrEmpty(strings.Split(op.Name(), "/"))`). The context cause is wrapped with `%w`, so `errors.Is(err, context.Canceled)` still holds for callers.

## Implementation

All error exits after the `UpdateDatabaseDdl` operation is accepted (initial poll error, `ctx.Done()`, in-loop poll error, and the final `op.Metadata()` fetch) now route through `handleDdlWaitError`, which keeps two concerns deliberately separate:

- **Schema-cache invalidation is unconditional.** Once the operation is accepted, any error exit may leave the schema changed server-side — a canceled wait whose operation later completes, or a multi-statement batch that fails at statement *k* after statements 1..*k*-1 already committed. `IncrementSchemaGeneration()` is cheap and always safe, and the schema-cache TTL bounds residual staleness, so it is always called rather than trying to prove the schema is unchanged.
- **Error-message classification is conditional.** Only cancellation/deadline errors get replaced with the operation-ID-bearing `SHOW OPERATION` hint. Genuine DDL failures are returned unchanged.

Cancellation classification lives in `isCancellationError`, which checks both:

- the standard-library context sentinels via `errors.Is` (matched when the context error is wrapped), and
- the gRPC / Spanner status codes (`status.Code`, `spanner.ErrCode`) for `codes.Canceled` / `codes.DeadlineExceeded`.

The status-code check is necessary because when the context is canceled while a `GetOperation` poll RPC is in flight, grpc-go returns a plain status error (`codes.Canceled`, "context canceled") that does **not** wrap `context.Canceled`; an `errors.Is`-only check would miss it and the hint would silently not fire. Tradeoff: a DDL that genuinely fails server-side could in principle carry those codes and be misreported as a cancellation. This is accepted because the in-flight-cancellation window is only reachable through those status codes, and misclassification only changes the error text — invalidation happens regardless.

### Scope notes

- The async DDL path (`Feature.AsyncDDL`) returns immediately without waiting and already calls `IncrementSchemaGeneration()`, so it needs no change.
- `SHOW OPERATION ... SYNC` (`ShowOperationStatement.executeSyncMode`) has its own separate polling loop that also returns a bare context error on cancel. It is a different statement family (monitoring an existing operation rather than submitting DDL) and was intentionally left out of this surgical change.

## Verification

- `make check` passes (test + lint + fmt-check), exit code 0.
- `TestDdlCancellationError` covers the error-formatting helper: operation-ID extraction from a full operation name and from a bare ID, the `SHOW OPERATION '<id>'` hint text, and `errors.Is` preservation for both `context.Canceled` and `context.DeadlineExceeded`.
- `TestIsCancellationError` covers the classification: context sentinels (bare and wrapped), plain and wrapped gRPC status errors for `codes.Canceled` / `codes.DeadlineExceeded` (the in-flight-RPC case), and negative cases (`codes.InvalidArgument`, plain error, nil).
- The concrete `*adminapi.UpdateDatabaseDdlOperation`-dependent wrapper (`handleDdlWaitError`) is not cheaply mockable without emulator choreography, so its wiring is verified by code inspection; the load-bearing classification and error text are unit-tested above.

### DIFF
--- a/internal/mycli/execute_ddl.go
+++ b/internal/mycli/execute_ddl.go
@@ -109,7 +109,7 @@ func executeDdlStatements(ctx context.Context, session *Session, ddls []string) 
 	metadata, err := pollDdl()
 	if err != nil {
 		teardown()
-		return nil, err
+		return nil, handleDdlWaitError(session, op, err)
 	}
 
 	if metadata != nil && bars != nil {
@@ -133,13 +133,13 @@ func executeDdlStatements(ctx context.Context, session *Session, ddls []string) 
 			// continue
 		case <-ctx.Done():
 			teardown()
-			return nil, ctx.Err()
+			return nil, handleDdlWaitError(session, op, ctx.Err())
 		}
 
 		metadata, err = pollDdl()
 		if err != nil {
 			teardown()
-			return nil, err
+			return nil, handleDdlWaitError(session, op, err)
 		}
 
 		if metadata != nil && bars != nil {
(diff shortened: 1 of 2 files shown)


apstndb (commented):

apstndb (commented):

PR #753, opened 2026-07-06
fix: propagate context and preserve gRPC status in InstanceExists

apstndb:
## Summary

Fixes several error-handling and context-propagation defects in `Session.InstanceExists` and neighbouring paths in `internal/mycli/session.go`. There is no linked issue; the defects were found by inspection against the current code.

### 1. `InstanceExists` ignored the caller's context (`session.go:607`)

`InstanceExists()` took no context and built its own with `context.WithTimeout(context.Background(), 30s)`. This meant the caller's cancellation (e.g. interactive interrupt, parent deadline) was ignored entirely.

Fix: `InstanceExists(ctx context.Context)` now layers the 30s timeout on top of the caller's ctx as an upper bound, so cancellation and any shorter parent deadline still propagate. The sole non-test caller is `NewAdminSession` (`session.go:413`), which already has a `ctx` in scope and now passes it through.

### 2. Method-2 fallback discarded the original error (`session.go:645`)

In the instance-admin-client-creation failure branch, the returned error read:

> `failed to create instance admin client: %v; original database list error: tried both spanner.databases.list and spanner.instances.get`

But the "original database list error" was never included — the `err` variable had been overwritten by the client-creation error, so the original `ListDatabases`/`dbIter.Next()` error was silently dropped and the message described an error it did not carry.

Fix: the original list error is now captured in a distinct variable (`listErr`) and both failures are surfaced with `errors.Join(listErr, err)` wrapped via `%w`, so the message now matches reality.

### 3. `%v` destroyed the gRPC status when wrapping (`session.go:670`, `session.go:699`)

`checking instance existence failed: %v` and `checking database existence failed: %v` flattened the underlying gRPC error into a plain string, so downstream `printError` could no longer extract the gRPC status code/details and degraded to a generic message.

Fix: both now wrap with `%w`, preserving the gRPC status through the error chain so `printError` can still unwrap it.

### 4. `RecreateClient` used `context.Background()` (`session.go:709`)

`RecreateClient` created its Spanner client with `context.Background()`. Its only caller is `executeStatement` (`cli.go:567`, the Aborted-transaction recovery path), which already has a live `ctx` in scope.

Fix: `RecreateClient(ctx context.Context)` now accepts and uses the caller's ctx. This is a surgical, single-caller signature change; no other paths call it, so nothing else churns.

## Why `%w` matters here

`printError` relies on unwrapping the error chain to recover the gRPC `status` (code + details). `%v` renders the error to a string and severs that chain, so a `NOT_FOUND`/`PERMISSION_DENIED` status becomes an opaque message. `%w` keeps the underlying error unwrappable so status-aware rendering keeps working.

## Behavior preserved

The two-method permission-fallback logic (databases.list first, then instances.get, with the specific NotFound / PermissionDenied / default handling for each) is unchanged. Only error wrapping, error preservation, and context plumbing changed.

## Verification

- Added `TestInstanceExists` (emulator-backed) covering the existing-instance, nonexistent-instance (NotFound), and cancelled-context paths. The cancelled-context case fails fast, which would not have been possible when the method used `context.Background()`.
- `make check` (test + lint + fmt-check) passes locally with exit code 0.

### DIFF
--- a/internal/mycli/cli.go
+++ b/internal/mycli/cli.go
@@ -564,7 +564,7 @@ func (c *Cli) executeStatement(ctx context.Context, stmt Statement, interactive 
 			// Once the transaction is aborted, the underlying session gains higher lock priority for the next transaction.
 			// This makes the result of subsequent transaction in spanner-cli inconsistent, so we recreate the client to replace
 			// the Cloud Spanner's session with new one to revert the lock priority of the session.
-			innerErr := c.SessionHandler.RecreateClient()
+			innerErr := c.SessionHandler.RecreateClient(ctx)
 			if innerErr != nil {
 				err = errors.Join(err, innerErr)
 			}
--- a/internal/mycli/session.go
+++ b/internal/mycli/session.go
@@ -631,6 +631,17 @@ func (s *Session) InstanceExists(ctx context.Context) (bool, error) {
 		return true, nil
 	}
 
+	// If the failure is caused by context cancellation or deadline (either the
+	// caller's ctx or the layered 30s bound), short-circuit and return that
+	// error directly. Method 2 would only issue another RPC on the same dead
+	// ctx and mislabel a cancellation as "checking instance existence failed".
+	if ctxErr := ctx.Err(); ctxErr != nil {
+		return false, ctxErr
+	}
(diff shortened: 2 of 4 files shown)


apstndb (commented):

apstndb (commented):

PR #754, opened 2026-07-06
fix: degrade SHOW PLAN NODE gracefully on malformed cached plans

apstndb:
## What and why

`SHOW PLAN NODE <id>` dumps the raw YAML of a cached plan node — a debugging aid whose whole point is inspecting whatever the last query left in the cache, however malformed.

Commit `77dd9a2` ("Show incoming parent links for plan nodes") added an incoming parent-link row to the output. To compute those links it calls `spannerplan.New(planNodes)` on the entire cached plan. That call can return an error on a malformed or unexpected plan shape (e.g. a child link whose `childIndex` is out of range, a `PlanNode.Index` that does not match its slice position, or a nil node). In the current Execute path the node YAML is already marshaled successfully *before* that call, but a `spannerplan.New` error now fails the **whole** statement.

That is a robustness regression: before `77dd9a2`, the raw YAML dump always succeeded regardless of overall plan integrity.

## The fix

When `spannerplan.New` (and therefore parent-link computation) fails, still return the node YAML and replace the parent-links row with a short note instead of failing:

```
parent links unavailable: spannerplan: childLink childIndex out of range: parent node 0 childLinks[0] has childIndex 99, len(planNodes)=2
```

The result still has two rows (node content + note) and `AffectedRows: 2`. The happy path is byte-identical: when the plan is well-formed, the parent-links row is produced exactly as before.

## Test changes

The existing `TestShowPlanNodeStatement_Execute` computed its expected node-content row by calling the **same** `protoyaml.Marshal(planNode, getGlobalOpts()...)` the production code uses. That expectation is tautological — it mirrors whatever the production marshaler emits, so it can never catch a marshaling regression.

- Replaced the computed expectation with literal golden YAML strings (a new `wantContent` field per case).
- Added a degraded-path case: a plan whose node 0 references a non-existent child index makes `spannerplan.New` fail; the test asserts the node YAML is still shown and the parent-links row is replaced by the `parent links unavailable: ...` note.

## Validation

`make check` passes (tests + `golangci-lint` clean + formatting).

### DIFF
--- a/internal/mycli/statements_explain_describe.go
+++ b/internal/mycli/statements_explain_describe.go
@@ -142,11 +142,18 @@ func (s *ShowPlanNodeStatement) Execute(ctx context.Context, session *Session) (
 		return nil, err
 	}
 
-	queryPlan, err := spannerplan.New(planNodes)
-	if err != nil {
-		return nil, err
+	// Computing incoming parent links requires reconstructing the whole plan,
+	// which can fail on a malformed or unexpected cached plan shape. SHOW PLAN
+	// NODE is a debugging command whose primary value is dumping the raw node
+	// YAML (already marshaled above), so degrade gracefully here: keep showing
+	// the node content and replace the parent-links row with a short note
+	// instead of failing the whole statement.
+	var parentLinksInfo string
+	if queryPlan, err := spannerplan.New(planNodes); err != nil {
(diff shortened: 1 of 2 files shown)


apstndb (commented):

apstndb (commented):
Rechecked current head after merging main; no blocking findings.

PR #755, opened 2026-07-06
test(lint_plan): add unit tests for CLI_LINT_PLAN linter (0% -> 100% coverage)

apstndb:
## Why

`internal/mycli/lint_plan.go` implements the `CLI_LINT_PLAN` heuristic query
plan linter — a documented feature (see `docs/query_plan.md`, "Query plan linter
(EARLY EXPERIMENTAL)"). Despite being user-facing, it had **0% test coverage**
and zero test references, so every heuristic and the graceful-degradation path
were unverified.

This PR adds table-driven unit tests that map each heuristic to at least one
plan that triggers it, and asserts on the exact lint message content. Expected
messages are hardcoded rather than produced by the formatting code under test,
so the assertions genuinely pin the documented behavior instead of being
tautological. Tests are unit-only (`go test -short`): the linter operates purely
on plan JSON, no emulator required.

## What each test covers

`TestLintPlan` (table-driven) — one case per heuristic in `lintPlan`:

- **Filter operator** (`filter.input.json`) — `"Filter"` display name.
- **Hash Join** + **Full scan** (`hash_join.input.json`) — `"Hash Join"` display
  name plus two nodes with `"Full scan": "true"` metadata.
- **Residual Condition** + **Full scan** (`scalar_subquery_with_filter_scan.input.json`)
  — `Residual Condition` child-link type plus a full-scan Table Scan.
- **Full scan in isolation** (`array_subqueries.input.json`) — single triggering
  node, exercising the full-scan rule on its own.
- **Sort** (`sort.input.json`, new) — `"Sort"` display name, including the key
  order rendered from the `Key` child link. The key `$SingerId` is resolved
  through the Scan node's variable definition, exercising the variable-lookup
  recursion in `lookupVar`.
- **Minor Sort** (`minor_sort.input.json`, new) — `"Minor Sort"` display name,
  matched before the plain `Sort` branch, reporting major/minor keys.
- **Hash iterator** (`hash_aggregate.input.json`, new) — `iterator_type=Hash`
  metadata on an Aggregate node.
- **No findings** (`no_findings.input.json`, new) — a plain (non-full) Table Scan
  under Serialize Result triggers nothing; asserts empty output.

`TestLintPlan_InvalidPlan` — a `PlanNode` whose `Index` does not match its slice
position makes `spannerplan.New` fail; verifies the linter logs and returns no
findings (linting is only informative) rather than erroring or panicking.

### Fixtures

Existing `testdata/plans/*.input.json` fixtures are reused wherever they already
trigger a rule. Four minimal fixtures were added only for rules no existing
fixture exercised: `sort`, `minor_sort`, `hash_aggregate`, and `no_findings`.
Each is kept as small as possible while still triggering (or, for
`no_findings`, deliberately not triggering) the heuristic.

## Coverage (lint_plan.go)

Measured with
`go test -short -coverprofile=... ./internal/mycli/ && go tool cover -func | grep lint_plan.go`:

| Function | Before | After |
|---|---|---|
| `lookupVar` | 0.0% | 100.0% |
| `descToKeyElem` | 0.0% | 100.0% |
| `buildVariableToNodeMap` | 0.0% | 100.0% |
| `LinkTypePred` | 0.0% | 100.0% |
| `formatKeyElem` | 0.0% | 100.0% |
| `lintPlan` | 0.0% | 100.0% |
| `formatKeyElemForLinkType` | 0.0% | 100.0% |

The whole file goes from 0% to 100% function coverage.

## Docs observation (no behavior change)

`docs/query_plan.md` illustrates the linter with a single example output that
demonstrates only the **Residual Condition** and **Full scan** heuristics. The
linter additionally implements **Filter**, **Hash Join**, **Sort**, **Minor
Sort**, and **Hash-iterator** heuristics that the docs do not spell out. This is
under-documentation, not a docs/code contradiction, so no behavior was changed;
noting it here for visibility.

## Verification

- `make check` passes (test + lint + fmt-check), exit 0.

### DIFF
--- a/internal/mycli/lint_plan_test.go
+++ b/internal/mycli/lint_plan_test.go
@@ -1,0 +1,145 @@
+package mycli
+
+import (
+	"testing"
+
+	sppb "cloud.google.com/go/spanner/apiv1/spannerpb"
+	"github.com/google/go-cmp/cmp"
+)
+
+// TestLintPlan exercises each heuristic implemented by lintPlan against a query
+// plan that triggers it. Expected messages are hardcoded (not produced by the
+// same formatting code under test) so the assertions verify the documented
+// linter behavior described in docs/query_plan.md. The output slice is a flat
+// list where each finding is a node-title header line followed by one or more
+// indented ("    ") message lines.
+func TestLintPlan(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		name string
+		// file is a plan fixture under testdata/plans. Existing fixtures are
+		// reused where they already trigger a rule; sort/minor_sort/
+		// hash_aggregate/no_findings are minimal fixtures added specifically
+		// for rules that no existing fixture exercised.
+		file string
+		// want is the complete expected lint output. nil means "no findings".
+		want []string
+	}{
+		{
+			// "Filter" display name -> Filter operator heuristic.
+			name: "filter_operator",
+			file: "testdata/plans/filter.input.json",
+			want: []string{
+				"1: Filter",
+				"    Potentially expensive operator Filter can't utilize index: Maybe better to modify to use Filter Scan with Seek Condition?",
+			},
+		},
+		{
+			// "Hash Join" display name -> Hash Join heuristic; the same plan
+			// also has "Full scan": "true" metadata on two scan nodes.
+			name: "hash_join_and_full_scan",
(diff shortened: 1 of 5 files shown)


apstndb (commented):

PR #756, opened 2026-07-06
test: de-flake integration sleeps and give the PTY test real assertions

apstndb:
## Summary

Two independent test-harness de-flaking fixes. Both target tests that
either waste wall-clock time on fixed sleeps or cannot catch regressions.

### 1. `integration_test.go` — stale-read test: fixed sleep → polling

`TestReadOnlyTransaction/begin ro with stale read` did a hard
`time.Sleep(10 * time.Second)` before an exact-staleness read.

- **What it waited for:** a `BEGIN RO 5` (exact staleness = 5s) read at
  `now - 5s` cannot observe the table until at least 5s have elapsed
  since the table was created by `initializeWithRandomDB`. The sleep was
  padding to guarantee the stale timestamp is valid.
- **Why the fixed sleep is bad:** it always burned 10s regardless of how
  quickly the emulator settled, and 10s is only loosely related to the
  real 5s staleness bound (over-provisioned "sleep for a while").
- **Fix:** poll the exact-staleness read (via a small `waitForCondition`
  helper, 100ms interval, 30s deadline) until `AffectedRows == 2` — i.e.
  the stale timestamp can see both initial rows. The test then proceeds
  immediately (typically ~5s) and fails with a clear message on deadline.
  Note: after the TypedRows migration a buffered SELECT reports its row
  count via `AffectedRows` (rows live in `Result.Typed`), so the poll
  checks `AffectedRows`, not `len(Result.Rows)`.

### 1b. `integration_test.go` — heartbeat test: sleep kept, now documented

`TestReadWriteTransaction/heartbeat: transaction is not aborted even if
the transaction is idle` also had `time.Sleep(10 * time.Second)`.

- **What it waits for:** it deliberately keeps the read-write transaction
  idle **longer than the server's ~10s idle timeout** to prove the
  background heartbeat keeps it alive.
- **Why it is left as a sleep:** this is genuinely required by
  server-side semantics — the value of the test is entirely in letting
  real wall-clock time exceed the idle timeout. Polling would observe
  "still alive" on every tick and prove nothing, and the wait cannot be
  shortened below the timeout it must exceed. The sleep now carries a
  comment explaining exactly this, instead of the terse original note.

### 2. `screen_width_vty_test.go` — real PTY assertions + hang guard

`TestDisplayResultWithPty` only did a single `ptmx.Read` and asserted
`n > 0`. That cannot catch rendering regressions, and a blocking
`ptmx.Read` on empty output could hang the test forever.

- **New assertions:** the expected rendered content is derived from the
  test inputs — `TableHeader.Render(false)` plus each cell value via
  `format.Texts` — and the test asserts every fragment appears in the PTY
  output. Assertions use substring containment because a PTY may translate
  `\n` to `\r\n` (ONLCR).
- **Hang guard:** the read runs in a goroutine bounded by a 5s timeout.
  On timeout the PTY is closed to unblock the pending `Read` and the test
  fails with a clear message. `*os.File.SetReadDeadline` is not supported
  on PTY masters on all platforms (it returns "file type does not support
  deadline" on macOS), so the goroutine+timeout approach is used instead
  of a read deadline.

## Stability evidence

All commands captured to files with exit codes checked explicitly.

- PTY test (`TestDisplayResultWithPty` + `TestGetTerminalSizeWithPty`):
  `-count=10` — pass.
- Affected integration tests (`TestReadOnlyTransaction` +
  `TestReadWriteTransaction`): `-count=5` — pass. Total wall time dropped
  from ~199s to ~54s across the 5 iterations.
- Per-subtest timing after the fix: stale-read subtest ~5s (was gated by
  the 10s sleep); heartbeat subtest ~10s (documented, unchanged).
- `make check` (full test suite against the emulator + lint +
  fmt-check): pass.

## Notes

No linked issue. No production code changes — tests only.

### DIFF
--- a/internal/mycli/integration_test.go
+++ b/internal/mycli/integration_test.go
@@ -1742,7 +1742,13 @@ func TestReadWriteTransaction(t *testing.T) {
 			t.Fatalf("unexpected error happened: %s", err)
 		}
 
-		// default transaction idle time is 10 secs
+		// This sleep is genuinely required by server-side semantics and cannot be
+		// replaced by condition polling. The point of the test is to keep the
+		// read-write transaction idle for longer than the server's ~10s idle
+		// timeout and prove the background heartbeat keeps it alive; the only way
+		// to exercise that is to let real wall-clock time pass. Polling would
+		// observe "still alive" immediately on every tick and prove nothing, and
+		// the wait cannot be shortened below the idle timeout it must exceed.
 		time.Sleep(10 * time.Second)
 
 		// second query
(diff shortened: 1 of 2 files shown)


apstndb (commented):

apstndb (commented):
Rechecked current head after merging main; no blocking findings.

Issue #695, opened 2026-07-04
READONLY guard bypassed by batch DDL / RUN BATCH DML / CREATE DATABASE (marker method name mismatch)


apstndb:
## Bug

The `MutationStatement` marker interface requires the unexported method `isMutationStatement()` (statement_processing.go), but three statement types define an exported `IsMutationStatement()` instead:

- `CreateDatabaseStatement` (statements.go)
- `BulkDdlStatement` (statements.go)
- `BatchDMLStatement` (statements.go)

They therefore silently fail the `stmt.(MutationStatement)` type assertion in `Session.ExecuteStatement`, which gates two behaviors: pending-transaction determination and the READONLY guard (`failStatementIfReadOnly`).

## Impact

With `SET READONLY = TRUE` (or `--readonly`):

- DDL typed interactively (a single `DdlStatement`) is correctly rejected, but the same DDL arriving as a batch (`BulkDdlStatement` - built by batch input files / `--execute` with multiple DDLs / RUN BATCH after START BATCH DDL) bypasses the guard and executes.
- `RUN BATCH` DML (`BatchDMLStatement`) and `CREATE DATABASE` similarly bypass the guard.

No compile-time assertions exist to catch marker-method typos, and the same fragile pattern is used by other markers (`isDetachedCompatible`, `isMetaCommand`).

## Fix plan

1. Rename the three methods to the unexported `isMutationStatement()`.
2. Add compile-time assertions (`var _ MutationStatement = (*BulkDdlStatement)(nil)` etc.) next to the interface definition for every implementing type, so a future typo fails the build instead of silently downgrading behavior.
3. Add regression tests: batch DDL / batch DML / CREATE DATABASE rejected in READONLY mode.

Found during the 2026-07 full-repo review.

PR #757, opened 2026-07-06
fix: block mutating CQL and ADD SPLIT POINTS in READONLY mode

apstndb:
## Summary

Closes two remaining gaps where a statement could mutate data while the session was in READONLY mode (`--readonly` / `SET READONLY=TRUE`). Both statements bypassed the single READONLY guard in `Session.ExecuteStatement` (`internal/mycli/session.go`), which only consulted the static `MutationStatement` marker.

## Gap 1: CQL statements bypassed the guard entirely

`CQLStatement` (`internal/mycli/statements.go`) does not implement `MutationStatement`, so the guard never saw it. As a result `CQL DELETE ...`, `CQL INSERT ...`, etc. executed and mutated data even under READONLY.

Unlike SQL statement types, CQL mutation-ness is not decidable from the Go type: it depends on the statement text. So a static marker is insufficient.

### Fix: fail-closed first-keyword classification + a per-statement guard marker

- New pure function `cqlStatementMutates(cql string) bool` classifies a CQL statement by its first whitespace-delimited keyword and is deliberately **fail-closed**: only `SELECT` is treated as read-only. Every other verb — `INSERT`/`UPDATE`/`DELETE`, `BATCH`, `TRUNCATE`, `CREATE`/`DROP`/`ALTER`, `GRANT`/`REVOKE` — and any unrecognized or empty keyword is treated as mutating so the guard blocks it. `SELECT` is the only clearly read-only data verb in the Spanner Cassandra adapter's supported surface; other Cassandra read verbs (e.g. `LIST`, `DESCRIBE`) are not part of that surface and are intentionally not allow-listed. Fail-closed means an unknown/future CQL verb can never silently bypass READONLY.
- New unexported marker interface `ConditionallyMutatingStatement { isConditionallyMutating() bool }` in `internal/mycli/statement_processing.go`, implemented by `CQLStatement`. It is consulted at the **same single guard call site** in `Session.ExecuteStatement`, preserving the central-guard property. Crucially, a conditionally-mutating statement participates only in the READONLY check and does **not** determine a pending Spanner transaction, because CQL does not use the Spanner transaction machinery.
- A compile-time assertion `var _ ConditionallyMutatingStatement = (*CQLStatement)(nil)` mirrors the `MutationStatement` assertion discipline from #695/#696: because the marker method is unexported, a typo would silently drop the type out of the interface and reopen the guard gap, and the assertion catches that at build time.

## Gap 2: ADD SPLIT POINTS executed unguarded

`AddSplitPointsStatement` (`internal/mycli/statements_split_points.go`) mutates database split state via the admin API but did not implement `MutationStatement`, so it ran unguarded under READONLY.

### Fix

Added `func (AddSplitPointsStatement) isMutationStatement() {}` plus the matching compile-time assertion in `statement_processing.go`, consistent with the other DDL-ish statements.

## Tests

Added to `internal/mycli/readonly_guard_test.go`:

- `TestCQLStatementMutates` — exhaustive table over the classifier: `SELECT` (incl. lowercase / leading whitespace) is read-only; `INSERT`/`UPDATE`/`DELETE`/`BEGIN BATCH`/`TRUNCATE`/`CREATE`/`DROP`/`ALTER`/`GRANT`/`REVOKE`, an unrecognized verb, empty, and whitespace-only are all mutating (fail-closed).
- `TestReadOnlyGuardBlocksMutatingCQL` — mutating CQL (`DELETE`, `INSERT`, unrecognized keyword) is rejected with `errReadOnly` at `Session.ExecuteStatement`. These short-circuit at the guard before `Execute`, so no Cassandra adapter is needed.
- `TestReadOnlyGuardAllowsReadOnlyCQL` — asserts `SELECT` CQL is not a static `MutationStatement` and `isConditionallyMutating()` returns false, so the guard lets it through. This checks the guard input directly rather than running `Execute` (a passing SELECT would attempt a real Cassandra/Spanner connection), keeping the test hermetic per the no-heavy-CQL-infra constraint.
- Extended `TestReadOnlyGuardCoversBatchStatements` with an `ADD SPLIT POINTS` case.

`make check` passes (test + lint + fmt-check).

## Release note

This is a behavior change worth a release-note entry: **READONLY mode now also blocks mutating CQL statements and `ADD SPLIT POINTS`** (previously allowed by omission). Read-only `CQL SELECT` remains allowed. No in-repo docs enumerate the set of statements READONLY blocks, so no README/docs update was required.

### DIFF
--- a/internal/mycli/readonly_guard_test.go
+++ b/internal/mycli/readonly_guard_test.go
@@ -28,6 +28,7 @@ func TestReadOnlyGuardCoversBatchStatements(t *testing.T) {
 		{desc: "batch DML", stmt: &BatchDMLStatement{DMLs: []spanner.Statement{spanner.NewStatement("UPDATE t SET id = 1 WHERE TRUE")}}},
 		{desc: "CREATE DATABASE", stmt: &CreateDatabaseStatement{CreateStatement: "CREATE DATABASE d"}},
 		{desc: "SYNC PROTO BUNDLE", stmt: &SyncProtoStatement{UpsertPaths: []string{"examples.ProtoType"}}},
+		{desc: "ADD SPLIT POINTS", stmt: &AddSplitPointsStatement{}},
 	} {
 		t.Run(tt.desc, func(t *testing.T) {
 			t.Parallel()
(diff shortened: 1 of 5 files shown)


apstndb (commented):

PR #758, opened 2026-07-06
test: fold build-tag zoo into the testing.Short() convention

apstndb:
## Summary

Fold the remaining build-tag zoo in the test suite into the single documented convention: `testing.Short()`. After this change, no test file carries a build tag; tier selection is done purely at runtime via `testing.Short()`, so every test compiles and runs in `make test`/CI and self-skips under `-short`.

## Background / history

Two mechanisms had accreted on top of the `testing.Short()` convention:

- **`//go:build integration`** on `integration_meta_command_test.go` and `integration_source_echo_test.go` (~775 lines). These files were **never executed** by any `make` target or CI job. They were only *compile-checked* by a later-added vet step (`go vet -tags integration,skip_slow_test ./...`). Because they never ran, they silently rotted against product changes.
- **`//go:build !skip_slow_test`** on `main_slow_test.go` and `session_slow_test.go`. This is a negative gate against a tag (`skip_slow_test`) that **nothing ever defines**, so these files already built and ran by default. The gate was dead — a no-op that only served the vet step above.

### Deleted: the two `integration`-tagged files (obsolete)

When re-enabled, these tests fail wholesale due to two intentional product changes since they last ran:

1. **Interactive mode now requires a real TTY.** They drive `cli.RunInteractive` with `bytes.Buffer` input/output; `initializeMultilineEditor` now errors `cannot run in interactive mode: stdout is not a terminal` because there is no TTY stream. Reviving them would require a PTY harness solely to re-test behavior already covered at the handler level.
2. **`NewCli` now performs an instance-existence RPC.** Their `createTestCli` helper builds a session with an empty connection, so `NewCli` fails with `checking instance existence failed ... Invalid project ID`. (Notably, the `\u`/use-database subtests failed at *construction* — they never actually exercised the use-database path, so they provided no real coverage to lose.)

Rather than build a PTY harness to duplicate existing coverage, both files are deleted. Every behavior they asserted is already covered by faster, working, non-obsolete tests:

| Deleted subtest(s) | Existing coverage |
|---|---|
| shell `\!` enabled / disabled / exit-status | `TestShellMetaCommand_Execute` (meta_commands_test.go) |
| prompt `\R` set / percent expansion | `TestPromptMetaCommand_Execute` + `TestParseMetaCommand` |
| tee `\T` new file / existing file / directory error | `TestTeeOutputMetaCommand_Execute` |
| tee `\T` quoted filename (parsing) | `TestParseMetaCommand` (`tee output with quotes`) |
| disable tee `\t` | `TestDisableTeeMetaCommand_Execute` |
| source `\.` parsing / quotes | `TestParseMetaCommand` (`source command with quotes`) |
| source `\.` file execution / syntax error / meta-in-file | `TestCli_executeSourceFile` |
| source `\.` non-existent file | `TestCli_executeSourceFile_NonExistentFile` |
| use database `\u` parsing / backticks | `TestParseMetaCommand` (`use database with backticks`) |
| meta command rejected in batch mode | `TestBuildCommands`/`cli_test.go` (`meta command at start of input`, `meta command after SQL statement`) + `TestCli_executeSourceFile` (`Meta command in file`) |
| source with ECHO enabled / disabled | `TestExecuteStatement_decorationsPrecedeStreamedRows` and other `EchoInput` cases (result_sink_test.go, cli_test.go) |

The two `integration`-tagged subtests that "passed" did so **vacuously**: `RunInteractive` bailed out at the TTY check before any statement ran, so their negative assertions (output does *not* contain X) were trivially satisfied.

### Removed dead `!skip_slow_test` gate

Dropped the tag line from `main_slow_test.go` and `session_slow_test.go`. Both already self-skip via `testing.Short()`; they already ran by default, so this is a behavioral no-op that just removes the dead gate.

### CI workflow

Removed the now-obsolete `vet tag-gated test files` step (and its comment) from `.github/workflows/run-tests.yaml`. Its only purpose was to compile-check the tag-gated files; with no tags left, `go vet` in the normal jobs covers everything.

### Docs

Updated `dev-docs/patterns/testing.md` to describe `testing.Short()` as the single tier-selection mechanism and to drop the description of the legacy tags.

## Wall-time impact

**Essentially zero.** This PR does not enable any previously-skipped work:

- The deleted `integration` files never ran under any `make`/CI target, so removing them subtracts nothing from `make test`.
- The `!skip_slow_test` files already ran by default, so removing their dead tag changes nothing.

Local `make check` (full suite incl. emulator) after the change: `internal/mycli` package ~34s, whole `make check` wall-clock ~41s — unchanged from before.

## Verification

- `go vet ./...` — clean.
- `make check` (test + lint + fmt-check, full non-`-short` suite against the Dockerized emulator) — passes, exit 0.


dev-docs/patterns/testing.md
```diff
@@ -14,12 +14,12 @@ The suite has two tiers, selected with the standard `-short` flag:
...
-  `runStatementTests` helper). Do NOT introduce new build tags for this.
-  Legacy tags remain in a few files: `//go:build integration` excludes
-  those files from default `go test` (CI compile-checks them with
-  `go vet -tags integration,skip_slow_test`). Files marked
-  `//go:build !skip_slow_test` are built by default and excluded only when
-  the `skip_slow_test` tag is defined (as in that CI vet step).
+  `runStatementTests` helper). This is the ONLY tier-selection mechanism:
+  `testing.Short()` is checked at runtime, so every test compiles and runs in
+  the default `go test`/`make test`, and self-skips under `-short`. Do NOT
+  introduce build tags (such as the former `integration` / `skip_slow_test`
```
apstndb, line 20: internal/mycli/integration_dump_test.go still starts with //go:build !short, so this new "ONLY tier-selection mechanism" statement is not true after the PR. Please either migrate that file by removing the tag (it already self-skips with testing.Short()) or narrow the docs and keep an explicit compile path for any remaining tag-gated tests.

apstndb (commented):
Found one blocking inconsistency in the testing-tier cleanup.

dev-docs/patterns/testing.md
apstndb: Good catch, thank you. Removed the dead `//go:build !short` line from `internal/mycli/integration_dump_test.go` in da756bf. Every test in that file already self-skips via `testing.Short()`, so the tag was a redundant no-op (nothing defines a `short` build tag). I also re-grepped all `*_test.go` files: no `//go:build` tier-selection tags remain anywhere, so the testing.md claim is now true repo-wide. `make check` passes.

Fixed in commit da756bf.

apstndb (commented):
Rechecked current head; the previous build-tag finding is fixed, and I found no further blocking issues.

PR #759, opened 2026-07-06
build(deps): upgrade ptyhelp to v0.2.3 and modernize docs-update

apstndb:
## Summary

Upgrade `github.com/apstndb/ptyhelp` from v0.2.1 to v0.2.3 (a `tool` directive dependency) and modernize the `docs-update` Makefile target to use its new pipe-mode + fence capabilities, replacing the two hand-rolled `awk` marker-splice pipelines.

## What v0.2.2/v0.2.3 bring (relevant items)

- **`patch` uses plain pipes by default** unless `-pty`/`-cols`/`-rows` is given. The `readme-help` block passes `-cols 120` (go-flags reads terminal width via `ioctl(TIOCGWINSZ)`), so it stays PTY-backed. The `statement-help` and `sysvars-help` blocks do not need a PTY and now run in plain-pipe mode.
- **New `-fence` flag** (default `text` wraps output in a ` ```text ` fenced block; `-fence=none` inserts raw Markdown). The `readme-help` block is ` ```text `-fenced and keeps the default. The `statement-help` and `sysvars-help` blocks are **raw Markdown tables**, so they use `-fence=none`.
- **`-normalize-eol` now normalizes the entire target file** (previously only captured stdout). This is a no-op here because the repo is LF-only; verified no file changes.
- **Hardening**: the target file is left untouched when the child exits non-zero, duplicate begin markers are rejected, marker matching is exact/trimmed, writes are atomic, and file mode is preserved.
- **New**: `-check` (exit 1 when the target would change — CI staleness check), `-dry-run`, `-timeout`, stdin (`-`) as a content source, and `ptyhelp version`.

## Why pipe-mode + `-fence=none` replaces the awk pipelines

The previous `awk` pipelines silently spliced captured stdout between the marker comments and `mv`'d a temp file over the target. That approach:

- did not detect duplicate/mismatched markers (a stray marker would silently corrupt the splice),
- performed a non-atomic capture-then-splice through a temp file and `mv`.

(Note on failure handling: because the capture and the splice were separate make recipe lines, a failing `go run` already stopped make before the awk splice ran, so the old approach did guard the common failure case. `ptyhelp patch` is at least equivalent and stricter: the check and the write are a single operation that aborts without rewriting when the child exits non-zero.)

`ptyhelp patch` also rejects duplicate begin markers, matches markers exactly, and writes atomically while preserving file mode. This removes ~9 lines of awk/temp-file plumbing in favor of two direct invocations. The informational `tmp/*.txt` outputs are preserved via the `-o` flag.

## Byte-identity verification

`README.md` and `docs/system_variables.md` are **byte-identical** before and after this change:

1. After the version bump alone (awk pipelines still in place): `git diff --exit-code -- README.md docs/system_variables.md` → clean.
2. After switching both blocks to pipe-mode `ptyhelp patch -fence=none`: `git diff --exit-code -- README.md docs/system_variables.md` → clean.
3. `ptyhelp patch -check` on the regenerated blocks exits 0 (no pending change).

`make check` passes (test + lint + fmt-check, exit 0). The `readme-sync` CI workflow runs `make docs-update` + `git diff --exit-code` and enforces the same invariant.

## CI staleness check

The `readme-sync` workflow already performs `make docs-update` followed by `git diff --exit-code -- README.md docs/system_variables.md`, which is equivalent to the new `-check` flag. A separate `docs-check` target would duplicate that gate, so the CI workflow is intentionally left unchanged.

## Transitive dependency change

ptyhelp v0.2.3 migrated its PTY backend, so `go mod tidy` swapped `github.com/aymanbagabas/go-pty` + `github.com/u-root/u-root` for `github.com/charmbracelet/x/conpty`.


apstndb (commented):

PR #766, opened 2026-07-06
fix: harden file-input safety for credentials, stdin, and SafeReadFile

apstndb:
## Summary

The CLI already applies the `filesafety` contract (size caps and file-type
checks) to SQL files (`config.go`, `cli.go`) and sample-database files
(`sample_databases.go`), but two file-like inputs bypassed it and one
`filesafety` helper had a stat-then-reopen race. This PR closes all three gaps so
size/type limits are enforced consistently across every file-like input, and
tightens `SafeReadFile` to validate the exact descriptor it reads from.

### 1. Credential file read was unbounded and bypassed filesafety

`readCredentialFile` (`internal/mycli/config.go:699-706`) opened the
`--credential` file with `os.Open` and read it with an unbounded `io.ReadAll`,
skipping the size/type checks used for other file inputs. It now routes through
`filesafety.SafeReadFile` with a 4 MiB cap (`FileSafetyOptions.MaxSize`).
Service-account key files are a few KiB, so 4 MiB is a generous ceiling that
still bounds the read and rejects directories/special files with clear errors.

### 2. stdin batch input was unbounded

`determineInputAndMode` (`internal/mycli/config.go:735` and `:757`) read
`--file=-` and non-TTY stdin with unbounded `io.ReadAll`, so a hostile or
accidental stream could exhaust memory. Both sites now use a shared
`readStdinCapped` helper that wraps stdin in `io.LimitReader` at the
`filesafety.DefaultMaxFileSize` limit and returns a clear error naming the limit
when exceeded, instead of silently truncating. Interactive mode is untouched.

### 3. SafeReadFile stat-then-reopen TOCTOU

`filesafety.SafeReadFile` (`internal/mycli/filesafety/file_safety.go:97-122`)
stat-ed the path, validated the resulting `FileInfo`, then re-opened the path
(`os.ReadFile` for regular files) — a time-of-check/time-of-use gap where the
path could be swapped between the stat and the read. It is now restructured to
validate the descriptor it actually reads from, mirroring
`streamio.openOutputFile`'s open-then-fstat pattern:

1. A pre-open `os.Stat` rejects unsafe targets before opening. This is kept
   because opening a FIFO `O_RDONLY` blocks until a writer appears, so we must
   not open one we are only going to reject.
2. After `os.Open`, `f.Stat()` re-validates against the opened descriptor; a
   path swapped after the initial stat (e.g. regular file replaced by a device
   or FIFO) is caught here, and the read comes from exactly the validated
   descriptor.
3. The read is capped via `io.LimitReader(f, max+1)` for BOTH regular and
   allowed non-regular inputs. Stat sizes are meaningless for pipes and can be
   stale for regular files, so the limit is enforced at read time as well.

`ValidateFileSafety` remains exported (used directly by tests) and existing
error-message semantics are preserved.

## Why consistency matters

The CLI treats untrusted file paths uniformly for SQL and sample data by routing
them through `filesafety`. Credentials and stdin were the two remaining
file-like inputs that skipped those guarantees; aligning them removes
inconsistent, unbounded reads from the codebase.

## Tests

- `filesafety`: added open-once read-cap coverage for regular files (at limit
  and over limit); existing FIFO (within/exceeds limit) and directory cases
  continue to cover non-regular and TOCTOU-adjacent behavior.
- `mycli`: added a credential-cap test (oversized file rejected with a
  too-large error) and unit tests for the stdin cap (within/at/over limit) via
  the testable `readStdinCapped` helper.

## Validation

`make check` (test + lint + fmt-check) passes.

### DIFF
--- a/internal/mycli/filesafety/file_safety.go
+++ b/internal/mycli/filesafety/file_safety.go
@@ -89,32 +89,51 @@ func ValidateFileSafety(fi os.FileInfo, path string, opts *FileSafetyOptions) er
 	return nil
 }
 
-// SafeReadFile reads a file after performing safety checks. For regular files
-// the size limit is enforced up front from Stat. Non-regular inputs (only
-// reachable with AllowNonRegular, e.g. process substitution FIFOs) report a
-// meaningless Stat size, so the read itself is capped at the size limit and
-// exceeding it is an error rather than a truncation.
+// SafeReadFile reads a file after performing safety checks, validating the same
+// file descriptor it reads from to avoid a stat-then-reopen TOCTOU.
+//
+// The flow mirrors streamio.openOutputFile's open-then-fstat pattern:
+//  1. A pre-open Stat rejects unsafe targets (e.g. non-regular files when
+//     AllowNonRegular is unset, or directories) before we open them. This
+//     matters because opening a FIFO O_RDONLY blocks until a writer appears, so
+//     we must not open one we are only going to reject.
+//  2. After os.Open, f.Stat() re-validates against the descriptor actually
+//     opened. If the path was swapped between the two stats (e.g. a regular
+//     file replaced by a device or FIFO), this catches it and we never read
+//     from an unexpected target.
+//
+// The read itself is always capped via io.LimitReader for BOTH regular and
+// allowed non-regular inputs. Stat sizes are meaningless for pipes and can be
+// stale for regular files, so the limit is enforced at read time rather than
+// only from the FileInfo size.
 func SafeReadFile(path string, opts *FileSafetyOptions) ([]byte, error) {
+	maxSize := maxFileSize(opts)
+
+	// Pre-open validation: reject unsafe targets before opening so we never
+	// block opening (e.g. a FIFO) a file we would only reject anyway.
 	fi, err := os.Stat(path)
 	if err != nil {
 		return nil, fmt.Errorf("failed to stat file %s: %w", path, err)
 	}
-
 	if err := ValidateFileSafety(fi, path, opts); err != nil {
 		return nil, err
 	}
-
-	if fi.Mode().IsRegular() {
-		return os.ReadFile(path)
-	}
-
-	maxSize := maxFileSize(opts)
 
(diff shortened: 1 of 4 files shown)


internal/mycli/filesafety/file_safety.go
```diff
@@ -89,33 +89,52 @@ func ValidateFileSafety(fi os.FileInfo, path string, opts *FileSafetyOptions) er
...
 		return nil, err
 	}
 
-	if fi.Mode().IsRegular() {
-		return os.ReadFile(path)
-	}
-
-	maxSize := maxFileSize(opts)
-
 	f, err := os.Open(path)
```
apstndb, line 129: Non-blocking nit: this still leaves a regular-file-to-FIFO race between the pre-open `os.Stat` and this blocking `os.Open`. The descriptor `f.Stat()` protects the bytes that get read after open succeeds, but it cannot run if `os.Open` blocks waiting for a FIFO writer. If that adversarial case is in scope, this needs a nonblocking open path; otherwise please soften the TOCTOU comment so it does not imply all FIFO swaps are caught before blocking.

apstndb (commented):
APPROVE-WITH-NITS: no blocking findings. The credential and stdin caps are fail-closed, and descriptor revalidation protects the bytes read from swapped paths. I left one non-blocking nit that the TOCTOU comment should not imply regular-to-FIFO swaps cannot still block at open.

internal/mycli/filesafety/file_safety.go
apstndb: Good catch. Softened the SafeReadFile doc comment to state the guarantee scope precisely: the fd re-validation ensures we never read from a swapped/unexpected file type, but a regular-file-to-FIFO swap between the pre-open Stat and os.Open can still block the blocking open until a writer appears. Documented as an accepted local-CLI limitation, with a note that O_NONBLOCK is deliberately avoided to keep process-substitution FIFOs working. Comment-only change in d134972.

Fixed in commit d134972.

PR #767, opened 2026-07-06
fix: enforce SET policy on ADD and correct generated sysvar docs

apstndb:
## Summary

Three related correctness fixes in the system-variable registry and its generated documentation. Each was independently verified against the varDef table and the docs generator.

### 1. `SET X += ...` bypassed the policy guards that `SET X = ...` enforces (fail-closed)

`VarRegistry.Set` (`internal/mycli/var_registry.go`) enforces mutation policy centrally from the def's metadata: read-only (`!def.settable()`), session-init-only (`initOnly`), and transaction-guarded (`txnGuard`). `VarRegistry.Add` — the `SET X += ...` path (`SetAddStatement` → `AddFromGoogleSQL` → `addFrom` → `Registry.Add`) — ran none of these checks and went straight to the ADD handler. That is a latent policy bypass: a variable a plain `SET` refuses to mutate could still be mutated via `+=`.

Fix: extract the shared check into `VarRegistry.checkSetPolicy(def)` and call it from both `Set` and `Add`, so the two mutation entry points cannot diverge (fail-closed by construction).

`SET LOCAL` interaction: there is no `SET LOCAL X += ...` grammar — the `SET LOCAL` pattern in `client_side_statement_def.go` only accepts `=`, and `SetLocalStatement` mutates through `SetFromGoogleSQL` (the Set path), which already honors `localAllowed()`. So there is no ADD path for `localAllowed()` to guard; the asymmetry was strictly Set-vs-Add.

Added table-driven tests (`TestRegistryAddEnforcesSetPolicy`) using small fake defs for readOnly+bindAdd, initOnly+bindAdd, and txnGuard+bindAdd, asserting the ADD handler never runs when a guard fires, plus `TestRegistryAddAllowedWhenSettable` to confirm the guard does not over-block.

### 2. Unimplemented variables were advertised as `read,write` in generated docs

`AUTOCOMMIT` and `RETRY_ABORTS_INTERNALLY` are bound to `UnimplementedVar` in `var_defs.go` (both `Get` and `Set` return unimplemented errors), yet `docs/system_variables.md` listed them with operations `read,write`. Drift CI cannot catch this: the generator output matched the checked-in file — the generator's *inputs* were stale, not the file.

Fix: `VarRegistry.ListVariableInfo` now derives an `Unimplemented` flag from the bound handler type (`rv.v.(*UnimplementedVar)`), not a hardcoded name list, so the status stays honest as variables are added or implemented. `helpVariableRows` renders the operations column as `unimplemented` for those vars. This renderer is shared by `HELP VARIABLES` and the `--sysvars-help` docs generator, so both stay consistent.

### 3. `CLI_FORMAT` description omitted implemented formats

The `CLI_FORMAT` desc in `var_defs.go` listed only TABLE/…/CSV and omitted the implemented `JSONL`, `SQL_INSERT`, `SQL_INSERT_OR_IGNORE`, and `SQL_INSERT_OR_UPDATE` values from `enums.DisplayMode`. Updated the desc to enumerate the full value set. Spot-checked the other enum-typed vars (CLI_PARSE_MODE, CLI_TABLE_STREAMING, CLI_STYLED_OUTPUT, CLI_WIDTH_STRATEGY, READ_LOCK_MODE, EXPLAIN_FORMAT) against their enumer tables — no further drift found.

## Validation

- `make check` passes (test + lint + fmt-check).
- `make docs-update` regenerated `docs/system_variables.md`; the three rows now read `unimplemented` / full CLI_FORMAT value list.


apstndb (commented):
APPROVE: no actionable findings. The ADD path now shares SET policy before invoking the add handler, the unimplemented operation rendering matches the current direct UnimplementedVar bindings, and the regenerated system-variable docs reflect those registry states.

docs/system_variables.md
```diff
@@ -30,98 +30,98 @@ whether a variable can be read (`SHOW`), written (`SET`), or appended to
...
+|:----------------------------------|:---------------|:--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `AUTOCOMMIT`                      | unimplemented  | A boolean indicating whether or not the connection is in autocommit mode. The default is true.                                                                                                                                                                                                                                                                                                                                                                                                                            |
```
apstndb, line 37: Non-blocking docs nit: the generated table now emits the operation token `unimplemented`, but the prose above the table only defines `read`, `write`, and `add`. Please extend the legend to say that `unimplemented` means both SHOW and SET intentionally return the placeholder/unimplemented error, otherwise the new token is unexplained in the reference.

apstndb (commented):
APPROVE-WITH-NITS: I found one non-blocking docs nit after the first pass: the new `unimplemented` operations token should be explained in the reference prose. No blocking code findings.

docs/system_variables.md
apstndb: Good catch. Added a sentence to the hand-written legend (outside the generated markers) explaining the unimplemented token: such variables are recognized for Spanner JDBC connection-property compatibility but currently reject both SHOW and SET. README has no operations-column prose, so no other doc needed updating. Regenerated docs via make docs-update with no drift beyond this edit. Fixed in 9bedb73.

Fixed in commit 9bedb732e100af320286d8f9c6c435b041cc2824.

PR #780, opened 2026-07-07
refactor: move optional statement families to the end of the def table (part of #778)

apstndb:
Part of #778 (sequenced between the seam PR #779 and the first extraction PR).

## Summary
- Move the GEMINI/BIGQUERY/CQL statement definitions after the CLI control group, to the very end of the core client-side statement table.
- Regenerate the README statement-help block (`make docs-update`): the three rows move below HELP/EXIT — the intended one-time diff.

## Why
The Feature seam (#779) appends feature-contributed defs after the core table. With the optional families already last, extracting them into feature packages (PR1-PR3 of #778) re-appends them at the same position, so dispatch order, HELP output, and the generated statement help stay **byte-identical** across each extraction PR. This PR front-loads the only order change into one reviewed diff.

## Blast radius (all from the same table, all cosmetic)
Besides the README block, the same reorder is visible in the interactive `HELP;` output, `--statement-help`, and the Ctrl+T statement-name candidate order — the three optional families now list after HELP/EXIT everywhere. No dispatch outcome changes.

## Safety
- Dispatch is first-match-wins; the only relative-order change is between {GEMINI, BIGQUERY, CQL} and {HELP, HELP VARIABLES, EXIT}, whose pattern languages are disjoint (bare anchored words vs keyword + whitespace + remainder), so no input string changes its dispatch target.
- The #728 non-shadowing invariant test passes over the reordered table; def count and every pattern/handler are unchanged (pure permutation).
- `make check`, `make docs-update` + drift check pass.

## Validation
- `make check` (test + lint + fmt-check): pass
- `make docs-update` then `git diff`: only the expected 3-row reorder in README.md
- Independent adversarial review: verdict SAFE TO MERGE (dispatch invariance proven over all inputs, not just Syntax-derived examples)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01V28qKRVwUCkpT5pepTksgr

Issue #778, opened 2026-07-07
Quarantine optional statement families (GEMINI/CQL/BIGQUERY) behind an explicit Feature seam and ship a slim build variant


apstndb:
Design for quarantining the three optional statement families (GEMINI/LLM, CQL, BIGQUERY) out of the core `internal/mycli` package behind an explicit registration seam, and shipping a slim build variant that compiles them out. Verified against main at f23499f.

## 1. Motivation and evidence

Three optional statement families sit on the core import path of `internal/mycli`, each with dedicated `Session` fields and bridging machinery:

| feature | code (incl. tests) | unique deps | Session coupling |
|---|---|---|---|
| GEMINI/LLM | ~2,500 lines (statements_llm*.go + doc_cache.go + doc_embed.go + tests) | genai, genaischema, developerknowledge-go, spanner-docs-embed | `docCacheMu`/`docCache` + lazy `getOrCreateDocCache` (session.go:124-125, 512-546) |
| CQL | ~125 lines interleaved in statements.go:663-787 | gocql, go-spanner-cassandra | `cqlCluster`/`cqlSession` lazy fields (session.go:128-129) |
| BIGQUERY | ~300 lines + ~65 in session.go | cloud.google.com/go/bigquery | `bqCredential`/`bqClient`/`bqClientKey` + `bigQueryClient()` (session.go:137-139, 916-951) + #775's SessionHandler credential carry-over |

- Binary: 83 MB and drifting up (77 MB a month ago).
- Marginal linked size over a `cloud.google.com/go/spanner` baseline (arm64, default ldflags, symbol-reference probes — lower bounds): bigquery **+9.7 MB**, gocql+go-spanner-cassandra **+2.9 MB**, genai+genaischema+dkapi+spanner-docs-embed **+3.6 MB**. Slim variant expectation: roughly **−16 MB (~20%)** plus the feature code itself. (zap does NOT go away with CQL: core gRPC-interceptor logging uses it.)
- Every new feature re-learns the same lessons without structural help: BIGQUERY shipped without READONLY classification (fixed in #775), without typed export (BigQuery values export as display strings — tracked as the feature's debt), and with a credential lifecycle bug (#775).

## 2. Goals / non-goals

Goals:
1. `internal/mycli` stops importing feature-only deps; `Session` sheds per-feature fields.
2. One explicit registration seam ("Feature") through which a feature contributes statements, system variables, CLI flags, and session state.
[...]

PR #781, opened 2026-07-07
refactor: extract BIGQUERY into feature/bigquery (part of #778)

apstndb:
Part of #778 (PR1 in the migration plan). First extraction on top of the Feature seam (#779) and the def-table reorder (#780).

## Scope

Move the BIGQUERY statement family out of core `internal/mycli` into the new `internal/mycli/feature/bigquery` package (cloud dep aliased as `bq`), so core no longer imports `cloud.google.com/go/bigquery`.

- **Statements + classifier + formatter** moved: `BigQueryStatement`, the #775 fail-closed first-keyword classifier (`bigQueryStatementMutates` and helpers), and `formatBigQueryValue`. The statement embeds `mycli.MutationClassifier` (`Classify` wired in the constructor over its own SQL) and `mycli.MarksDetachedCompatible`; the `ConditionallyMutatingStatement` / `DetachedCompatible` compile-time assertions move with it. The core assertion for `BigQueryStatement` is removed.
- **Def** moved into `Feature()`. Each `Feature()` call allocates a **fresh feature-owned config** captured by the def closure, the `Variable` handlers, and constructed statements — no package-level mutable state (session-isolation commitment from the #779 review thread).
- **System variables** `CLI_BIGQUERY_PROJECT` / `CLI_BIGQUERY_LOCATION` / `CLI_BIGQUERY_MAX_BYTES_BILLED` become `FeatureVar`s with their descriptions moved **verbatim**; the nullable-int NULL display for `MAX_BYTES_BILLED` is preserved via `mycli.NullableIntVar`. The `BigQuery*` fields are removed from `FeatureVars`. The `CLI_BIGQUERY_PROJECT` -> `CLI_PROJECT` fallback stays in client-build code (via `Session.ProjectID()`).
- **Client cache** reimplemented on `mycli.FeatureState` (key `"bigquery.client"`) using `Session.AuthOptions`. It preserves the rebuild-on-project/location-change rule and closes the client at `Session.Close` via `io.Closer`. The old deliberately lock-free field cache is replaced by a mutex-guarded cache (atomic check-then-act, per styleguide Concurrency).

## Credential re-homing (deletes #775 carry-over machinery)

The raw `--credential` bytes get a durable home on the startup-config group (`StartupConfig.Credential`, read-only, not a registered variable), exposed via `Session.CredentialBytes()` (defensive copy). Because startup config is shared across USE/DETACH session replacement, the credential resolves the same way regardless of session switches **by construction**. This lets the PR delete:

- `SessionHandler.credential`, `applyBigQueryCredential`, `newSessionHandlerWithCredential` (restoring the simple `NewSessionHandler`);
- the `session.bqCredential` assignment in `createSession`, plus the `bqCredential`/`bqClient`/`bqClientKey` fields, `bigQueryClientKey`, `bigQueryClient()`, `createBigQueryClientOptions`, `bigQueryProject()`, and the `bqClient` close block in `Session.Close`.

The #775 credential-handoff regression is preserved, adapted to the new accessor (`TestSessionCredentialBytesFromDurableConfig`).

Note: the long-dead `Cli.Credential` field referenced in #778 section 4.6 no longer exists in the current tree, so there was nothing to delete there (see Deviations).

## Order-contract change + disclosed one-time diff

Features append **after** the core table, so extracting BIGQUERY first moves its statement-help row after CQL's. The final registration order is fixed as **(llm, cql, bigquery)** (not the pre-extraction core order llm, bigquery, cql), and `all.All()`'s doc comment is updated accordingly. This front-loads the single **2-row swap** (BIGQUERY and CQL) into this PR; PR2 (CQL) and PR3 (LLM) re-append at the same relative positions and stay byte-identical.

- `docs/system_variables.md`: **byte-identical** (variables are listed alphabetically; `ListVariableInfo` now also enumerates feature `varDef`s and `renderSystemVariablesHelp` is threaded with the feature set, so the generated reference keeps documenting the full binary's surface).
- `README.md`: exactly the disclosed BIGQUERY/CQL statement-help row swap, nothing else.

## Tests

- Feature package: classifier table, `formatBigQueryValue`, def dispatch via `mycli.BuildStatementWithDefs`, `FeatureVar` Get/Set semantics (incl. NULL display), fresh-per-`Feature()` isolation, and the project fallback.
- Dispatch-level READONLY guard (#775 regression) moves to an external `mycli_test` that registers `bigquery.Feature()` and drives the real guard through `Session.ExecuteStatement` (both blocks-mutating and allows-read-only, the latter proving the SELECT reaches Execute and fails on config, not READONLY). A small test-only bridge (`export_test.go`) exposes a READONLY test session and the READONLY-error predicate without widening the production API.
- The #728 invariant tests now iterate `MergedStatementDefs(all.All()...)`, so they keep covering BIGQUERY after extraction.
- `createBigQueryClientOptions`' emulator-auth-skip coverage is preserved as a core test over the exported `Session.AuthOptions(..., false)`.

## Validation

- `make check` (test + lint + fmt-check): pass
- `make check-race`: pass
- `go vet ./...`: clean
- `make docs-update` then `git diff`: `docs/system_variables.md` byte-identical; `README.md` shows only the BIGQUERY/CQL 2-row swap
- Dep hygiene: `go list -deps ./internal/mycli | grep cloud.google.com/go/bigquery` -> empty (root main still imports it via `feature/all`, which is correct)

## Tracked debt

The typed-export limitation (BigQuery values export as display strings, not typed cells) is recorded in the feature package doc comment.

## Deviations from the spec

- The `Cli.Credential` field named in #778 section 4.6 for deletion is not present in the current tree (already removed in an earlier change); no code change was needed for it.
- Two changes beyond the literal file list were required to hold the byte-identity gate, both consequences of moving the vars to feature registration: `ListVariableInfo` now iterates feature `varDef`s in addition to the core `varDefs` table, and `renderSystemVariablesHelp` is threaded with the feature set. Without these, the three BigQuery rows would drop out of `docs/system_variables.md` and `HELP VARIABLES` in the full binary.

Part of #778.

https://claude.ai/code/session_01V28qKRVwUCkpT5pepTksgr


apstndb (commented):
REQUEST-CHANGES: I found one blocking regression.

Finding:
- [P2] Feature variables disappear from `SHOW VARIABLES` and variable-name completion. `ListVariableInfo` was updated to include `r.sv.featureVarDefs`, so `HELP VARIABLES` and generated docs still see `CLI_BIGQUERY_PROJECT`, `CLI_BIGQUERY_LOCATION`, and `CLI_BIGQUERY_MAX_BYTES_BILLED`. But `VarRegistry.ListVariables` still iterates only the core `varDefs` table. `SHOW VARIABLES` calls `session.systemVariables.ListVariables()`, and fuzzy variable-name completion uses the same map, so after this extraction the full binary can `SET`/`SHOW VARIABLE` the BigQuery vars by exact name but no longer lists them in `SHOW VARIABLES` or completion. Those rows existed before the move because they were in core `varDefs`. Please make the variable enumeration path iterate feature var defs as well (and the analogous `ListMultiValues` path for consistency/future feature vars), then add regression coverage for `SHOW VARIABLES` or the registry listing with `bigquery.Feature()`.

References:
- `internal/mycli/var_registry.go:205-214`
- `internal/mycli/statements_system_variable.go:46-52`
- `internal/mycli/fuzzy_finder.go:846-853`

Validation:
- Clean merge into current `origin/main` (`3c5315727cf246c7e1c23533cb0b65342f788d07`) in a temporary clone.
- `go test ./internal/mycli ./internal/mycli/feature/bigquery -run 'TestFeature|TestBigQuery|TestFormatBigQueryValue|TestReadOnlyGuard.*BigQuery|TestClientSideStatementDefsNonShadowing|TestClientSideStatementDefsCompletionConsistency'`
- `go list -deps -f '{{if eq .ImportPath "cloud.google.com/go/bigquery"}}{{.ImportPath}}{{end}}' ./internal/mycli` produced no output, so the core package dependency split holds.

apstndb:
Addressed the REQUEST-CHANGES finding in abba75f.

- **[P2] Feature variables missing from `SHOW VARIABLES` / completion**: fixed at the root rather than per-method — `ListVariables`, `ListMultiValues`, and `ListVariableInfo` now all iterate through a single `VarRegistry.forEachDef` helper (core `varDefs` followed by `sv.featureVarDefs`), so no enumeration surface can silently drift from the others again. Regression coverage added in both suggested forms: `TestVariableListingIncludesBigQueryVars` (external package, real `bigquery.Feature()` registered, asserts the exact map SHOW VARIABLES and fuzzy completion consume) and `TestFeatureVarsAppearInEnumerationSurfaces` (internal, pins the mechanism for single-value and multi-value feature vars).
- Also folded in two non-blocking findings from the independent adversarial review: the stale `statements.go` comment pointing at the deleted `statements_bigquery.go`, and `MutationClassifier` now fails closed (classifies as mutating) on a nil `Classify` instead of panicking on a zero-value statement, with a test.

Validation: `make check` pass, `make docs-update` still produces zero diff (README/docs byte-identity holds), CI running.

apstndb (commented):
APPROVE: previous blocking finding is resolved; I found no new blocking issues.

Reviewed the delta from b6cdb60a64d631a7b47218e768ee7d11d8562aff to abba75ffe366997167b40c0a6f2ef8368f526211. The feature-variable listing fix routes `ListVariables`, `ListMultiValues`, and `ListVariableInfo` through the same core+feature def iterator, and the added tests cover both the mechanism and the real BIGQUERY feature variables behind SHOW VARIABLES / completion.

Validation:
- Clean merge into current `origin/main` (`3c5315727cf246c7e1c23533cb0b65342f788d07`) in a temporary clone.
- `go test ./internal/mycli ./internal/mycli/feature/bigquery -run 'TestVariableListingIncludesBigQueryVars|TestFeatureVarsAppearInEnumerationSurfaces|TestMutationClassifierDelegates|TestFeature|TestBigQuery|TestFormatBigQueryValue|TestReadOnlyGuard.*BigQuery|TestClientSideStatementDefsNonShadowing|TestClientSideStatementDefsCompletionConsistency'`
- `go list -deps -f '{{if eq .ImportPath "cloud.google.com/go/bigquery"}}{{.ImportPath}}{{end}}' ./internal/mycli` produced no output.

PR #782, opened 2026-07-07
refactor: extract CQL into feature/cql (part of #778)

apstndb:
## Summary

PR2 of #778: extract the CQL (Cassandra interface) statement family out of the core `internal/mycli` package into `internal/mycli/feature/cql`, behind the Feature registration seam. After this change core no longer imports `github.com/gocql/gocql` or `github.com/googleapis/go-spanner-cassandra`. Follows the BIGQUERY extraction pattern (#781).

## Scope

- **New package `internal/mycli/feature/cql`**: moves `CQLStatement`, the fail-closed `firstCQLKeyword`/`cqlStatementMutates` classifier, `formatCassandraTypeName`, and `Execute`. The `zapcore.WarnLevel.String()` usage moves with it (zap stays a core dep for gRPC-interceptor logging; only the gocql/spancql import lines leave core).
- **Marker wiring**: `CQLStatement` embeds `mycli.MutationClassifier` with `Classify` wired in `newCQLStatement` to `cqlStatementMutates` over its own CQL text (mirrors `newBigQueryStatement`), carries its own `var _ mycli.ConditionallyMutatingStatement = (*CQLStatement)(nil)` assertion, and the core assertion is removed. It is deliberately **not** `MarksDetachedCompatible` (the Cassandra adapter binds to the Spanner database, so CQL can't run in detached mode).
- **`Feature()`**: contributes the "CQL" def only (EARLY EXPERIMENTAL note preserved), no vars, no flags, fresh instances per call, no package-level mutable state.
- **Session surgery**: removed `cqlCluster`/`cqlSession` fields and the `cqlSession` Close block. The lazy CQL session is reimplemented on `mycli.FeatureState` (key `cql.session`) as a mutex-guarded, build-once `io.Closer` holder that closes the gocql session (the cluster is not separately closed, same as today), preserving today's `spancql.NewCluster` build (`DatabaseUri: session.DatabasePath()`, `Timeout = 5s`) and error messages exactly. No rebuild-on-key-change rule (DatabasePath can't change on a live Session).
- **Registration**: `feature/all.All()` returns `{cql.Feature(), bigquery.Feature()}`; merged order stays LLM, CQL, BIGQUERY (LLM still in core at the end of the core table).
- **Tests**: classifier (`TestCQLStatementMutates`) and dispatch/classification tests move into `feature/cql`; the dispatch-level READONLY guard for mutating CQL moves into an external `package mycli_test` file (`cql_readonly_guard_test.go`).

## Byte-identity proof

`make docs-update` then `git diff --exit-code README.md docs/system_variables.md` -> exit 0 (byte-identical). The CQL statement-help row keeps its position because the merged order (llm, cql, bigquery) matches the pre-extraction ordering already established by #780/#781.

## Dependency check

```
$ go list -deps ./internal/mycli | grep -E 'gocql|go-spanner-cassandra'
(no output)
```

Root main still pulls them via `feature/all` (correct):

```
$ go list -deps . | grep -E 'gocql|go-spanner-cassandra'
github.com/gocql/gocql/internal/lru
github.com/gocql/gocql/internal/murmur
github.com/gocql/gocql/internal/streams
github.com/gocql/gocql
github.com/googleapis/go-spanner-cassandra/logger
github.com/googleapis/go-spanner-cassandra/adapter
github.com/googleapis/go-spanner-cassandra/cassandra/gocql
```

## Validation

- `make check` — pass (0 lint issues, formatting correct)
- `make check-race` — pass
- `go vet ./...` — clean
- docs byte-identity — no diff

## Deviation from the task spec (allowed-SELECT guard test)

The task asked the external guard test to mirror `bigquery_readonly_guard_test.go` "exactly", including running an allowed `CQL SELECT` through `Session.ExecuteStatement` and expecting it to "fail later for a non-READONLY reason". That is **not** safe for CQL and is intentionally not done:

- BigQuery's `Execute` has a pre-network guard (`effectiveProject` returns `""` -> immediate "project not configured" error), so the allowed SELECT fails fast with no I/O.
- CQL's `Execute` has no such guard; a passing SELECT reaches `spancql.NewCluster`, which **panics** (`NewCluster` panics when `NewTCPProxy`/`newAdapterClient` fails) whenever the Cassandra adapter client cannot be built — i.e. always in unit tests, with or without ADC. Running it through `Execute` would crash the test binary, not return an error. This is the same reason the original in-core `TestReadOnlyGuardAllowsReadOnlyCQL` asserted classification directly instead of calling `Execute`.

Coverage is preserved without weakening the #757 guarantees:
- `cql_readonly_guard_test.go` runs the **blocked** mutating cases through real dispatch + `ExecuteStatement` (short-circuit at the guard, no adapter built) and asserts the READONLY sentinel; the allowed SELECT is asserted at dispatch level to be a non-static `MutationStatement`.
- `feature/cql/cql_test.go` (`TestCQLStatementConditionalMutation`) asserts the wired classifier directly (read-only SELECT -> non-mutating, DELETE -> mutating) via the exported `Classify` field.
- `feature/cql/classifier_test.go` (`TestCQLStatementMutates`) keeps the exhaustive fail-closed keyword coverage.

Part of #778

https://claude.ai/code/session_01V28qKRVwUCkpT5pepTksgr

### DIFF
--- a/internal/mycli/client_side_statement_def.go
+++ b/internal/mycli/client_side_statement_def.go
@@ -1191,10 +1191,11 @@ var clientSideStatementDefs = []*clientSideStatementDef{
 			return &ExitStatement{}, nil
 		},
 	},
-	// The optional statement families (GEMINI/CQL) sit at the end of the table.
-	// BIGQUERY was extracted into internal/mycli/feature/bigquery (#778); the
-	// remaining families follow in PR2/PR3. Feature-contributed defs are appended
-	// after this core table in feature/all.All() order (llm, cql, bigquery).
+	// The optional statement family GEMINI/LLM sits at the end of the table.
+	// BIGQUERY (#778) and CQL (#778) were extracted into internal/mycli/feature/
+	// {bigquery,cql}; GEMINI/LLM follows in PR3. Feature-contributed defs are
+	// appended after this core table in feature/all.All() order (llm, cql,
+	// bigquery).
 	// LLM
 	{
 		Descriptions: []clientSideStatementDescription{
@@ -1207,20 +1208,6 @@ var clientSideStatementDefs = []*clientSideStatementDef{
 		Pattern: regexp.MustCompile(`(?is)^GEMINI\s+(?P<text>.*)$`),
 		HandleGroups: func(groups map[string]string) (Statement, error) {
 			return &GeminiStatement{Text: unquoteString(groups["text"])}, nil
-		},
-	},
-	// Cassandra interface
-	{
-		Descriptions: []clientSideStatementDescription{
-			{
-				Usage:  `Execute CQL`,
-				Syntax: `CQL ...`,
-				Note:   "EARLY EXPERIMENTAL",
-			},
-		},
-		Pattern: regexp.MustCompile(`(?is)^CQL\s+(?P<cql>.+)$`),
-		HandleGroups: func(groups map[string]string) (Statement, error) {
-			return &CQLStatement{CQL: groups["cql"]}, nil
 		},
 	},
 }
--- a/internal/mycli/cql_readonly_guard_test.go
+++ b/internal/mycli/cql_readonly_guard_test.go
@@ -64,23 +64,39 @@ func TestReadOnlyGuardBlocksMutatingCQL(t *testing.T) {
 }
 
 // TestReadOnlyGuardAllowsReadOnlyCQL verifies, at the dispatch level, that a
-// read-only CQL SELECT is not classified as a static MutationStatement, so the
-// READONLY guard does not statically block it.
+// read-only CQL SELECT is not classified as a static MutationStatement and that
+// the DISPATCH-BUILT statement's conditional classifier reports non-mutating,
+// so the READONLY guard lets it through. Asserting the classification on the
+// statement produced by the real def table (not one built via the constructor)
+// catches a HandleGroups regression that returns a zero-value CQLStatement:
+// with the fail-closed nil-Classify default, that regression would silently
+// block CQL SELECT in READONLY mode (#782 review nit).
 //
 // Unlike the BigQuery counterpart, this does NOT run the SELECT through
 // Session.ExecuteStatement: a passing SELECT proceeds to CQLStatement.Execute,
 // which builds the Cassandra adapter via spancql.NewCluster. That call panics
 // when the adapter client cannot be constructed (no live Spanner backend in
 // tests), so it cannot "fail later for a non-READONLY reason" the way BigQuery's
-// project-not-configured guard does. The complementary proof that a wired SELECT
-// classifier reports non-mutating (so the conditional guard lets it through)
-// lives in feature/cql (TestCQLStatementConditionalMutation), which asserts the
-// classification directly without touching the adapter.
+// project-not-configured guard does.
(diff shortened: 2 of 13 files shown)


apstndb:
Internal review results (orchestrator line review + independent adversarial review): both conclude SAFE TO MERGE.

Verified under attack: classifier and formatCassandraTypeName byte-identical to the pre-move code; Execute identical modulo exported wrappers (including the empty-header `toTableHeader(headers)` vs `NewTableHeader(headers...)` equivalence — both return nil); READONLY coverage equal-or-better (mutating CQL now blocked through real dispatch, #757 keyword table preserved byte-identically, plus a new DELETE-classify assertion); docs zero-diff; core dependency graph free of gocql/go-spanner-cassandra; merged order LLM, CQL, BIGQUERY.

Two honest notes for the record, neither blocking:
1. Failure-retry nuance: the old code cached the Cassandra cluster before `CreateSession()`, so a transient `CreateSession` failure retried on the same cluster/proxy; the new holder rebuilds the cluster on retry. Unobservable in practice because `NewCluster` panics on proxy-setup failure before a failing `CreateSession` is reachable (go-spanner-cassandra v0.7.0, spanner.go:88-92, 116-120).
2. That `NewCluster` panic is PRE-EXISTING production behavior (the old in-core Execute called the same function): a CQL statement on a session where the adapter cannot be built panics the REPL. Recorded as a post-series follow-up candidate (recover in feature/cql or an upstream report) — same class as the parse-panic findings already queued.

apstndb (commented):
APPROVE-WITH-NIT: I found no blocking behavior issues in the CQL extraction.

One non-blocking test coverage gap is worth tightening before or after merge:

- [P3] `internal/mycli/cql_readonly_guard_test.go:79`: the dispatch-built SELECT allow-path test only asserts that the returned statement is not a static `MutationStatement`. The wired classifier is asserted in `feature/cql` through `newCQLStatement`, and the production `HandleGroups` currently does call that constructor, so the implementation is correct. Still, the dispatch test (or `TestFeatureDefDispatch`) should also assert that the built `*cql.CQLStatement` classifies `SELECT` as non-mutating. That would catch a future `HandleGroups` regression that accidentally returns a zero-value `CQLStatement` and causes READONLY mode to fail-closed for CQL SELECT.

Validation I ran:
- `go test ./internal/mycli ./internal/mycli/feature/cql -run 'TestCQL|TestReadOnlyGuard.*CQL|TestFeatureDefDispatch|TestClientSideStatementDefsNonShadowing|TestClientSideStatementDefsCompletionConsistency|TestFeature'`
- `go list -deps ./internal/mycli` contains no `gocql` / `go-spanner-cassandra` imports
- `go list -deps .` still contains the CQL dependencies through the full binary feature set

apstndb:
Addressed the APPROVE-WITH-NIT [P3] in 2bdf95d: `TestReadOnlyGuardAllowsReadOnlyCQL` now asserts the conditional classification on the DISPATCH-BUILT statement (via a test-only `ClassifyForTest` bridge, since the marker method is unexported): SELECT must classify non-mutating and DELETE mutating on statements produced by the real def table. A `HandleGroups` regression returning a zero-value `CQLStatement` (nil Classify fails closed) now surfaces as a test failure instead of silently blocking CQL SELECT in READONLY mode.

Issue #936, opened 2026-09-12
Raise full-suite statement coverage to 80% with behavior-focused tests


apstndb:
Raise statement coverage to at least 80% on integrated main while adding tests that protect observable behavior.

Baseline: `bc14121b2deb58a18549ae8658b7ba7935a7a438`. The existing `make test-coverage` succeeds locally and `go tool cover -func=tmp/coverage.out` reports **78.4%**. `make check` also passes. The local full suite uses the existing Colima Docker socket. Coverage is measured on the full suite, including emulator-backed tests; unit-only measurements are diagnostic.

## Acceptance criteria

- [ ] At least 80% in the existing CI coverage report and in the full-suite `go tool cover -func` summary on integrated main.
- [ ] Preserve package scope, generated-code exclusions, and full-suite execution; do not remove executable code or change exclusions to inflate coverage.
- [ ] Assert externally meaningful values, errors, state transitions, or resource cleanup. New tests must detect plausible regressions and use deterministic synchronization where needed.
- [ ] Keep test changes in focused PRs. Prefer existing in-process fakes and helpers; no real cloud credentials, broad production refactors, or new mocking frameworks.
- [ ] Run `make check` before every push. All hosted CI checks must pass, and the current head must receive independent review before squash merge.
- [ ] Once achieved, raise the existing coverage floor to 80% in a separate small PR and validate it against the integrated suite.
- [ ] Record the final main SHA, measured coverage, and related PRs here through PR completion comments; close through the final PR.

## Work allocation

Codex owns decomposition, review feedback and GitHub replies, final verification, and merging. Cursor Projects receives bounded implementation instructions directly; transient test tasks do not each require a child Issue. A separately actionable bug or design change should receive its own Issue rather than silently expanding a test PR.

Initial test lanes avoid files currently being refactored in open PRs #907 and #931–#935:

1. Query-profile decoding and display: `internal/mycli/statements_query_profile.go`, including malformed profile/statistics/plan data, empty results, formatting, and fake-transport statement execution where useful.
2. Offline CLI startup/help/error behavior in `internal/mycli/app.go`, using existing flag and output-capture helpers without depending on a real Spanner service.
3. BigQuery result-value formatting: nullable values, timestamp/civil values, numeric precision, composite values, and fallback/error behavior in `internal/mycli/feature/bigquery/format.go`.

[...]

PR #937, opened 2026-09-12
Test BigQuery nullable and temporal value formatting

apstndb:
Part of #936.

Expand value-formatting tests for nullable BigQuery values, distinguishing SQL NULL from valid zero, false and empty values. Add fixed expected output for temporal precision, NUMERIC/BIGNUMERIC scale, nil decimals, collection values and the JSON-encoding fallback. Production behavior is unchanged.

Validation: focused formatter tests and independent review passed; full `make check` passed on both the cloud worker and the integrated host tree. The combined coverage result is tracked in #936 with the existing measurement scope.


apstndb (commented):
Reviewed `8f5c3e39ceb3dec09cac6ec65b62ef160e82a4fd`: no actionable findings. The one-file test delta covers null versus valid zero/empty values, temporal precision, decimal scales and the existing encoding fallback with explicit expected output.

Focused tests, full integrated-host `make check`, and independent review passed. The final tree `563478ec8fae74a50d734e819f3d522ecb6c31d9` is identical to the tested integration tree and includes current main. CI remains the merge gate.

PR #939, opened 2026-09-12
Test offline CLI help and startup validation

apstndb:
Part of #936.

Test offline statement/system-variable help, sample listings, usage output destinations, and startup validation failures before a database runtime starts. A small output-writer entry point lets help tests capture output without replacing process-wide stdout; the existing entry point still supplies `os.Stdout`.

Embedded-runtime startup-error cases run sequentially because setup changes the testcontainers global logger. This fixes the race reproduced during review. Other help and usage tests retain parallel execution.

Validation: focused race tests, independent review and full integrated `make check` passed. The combined full-suite `go tool cover -func` summary is 80.7%, using the existing measurement scope and exclusions.


apstndb (commented):
Reviewed `a0ba258358fec7476c1f5864951355edc386ba54`: no actionable findings. Help/usage tests verify their output destination, and startup-error tests assert validation failures before connecting or starting an embedded runtime.

Review reproduced a race in parallel embedded-runtime setup. The startup-error parent and its descendants now run sequentially because that setup changes the process-wide testcontainers logger. Focused race tests and full host `make check` passed after the fix. Independent review accepted the small output-writer seam and the integration.

The final tree `b140136e1592c5fed03362e868692c2da4c67872` is identical to the tested tree, includes current main, and measured 80.7% with the unchanged full coverage target. CI remains the merge gate.

Issue #941, opened 2026-09-12
Raise test coverage to 85% while retaining the 80% CI floor


apstndb:
## Goal

Raise the existing full-suite statement coverage to at least 85% on main, while retaining the CI acceptance floor at 80%. The previous main CI result was 80.6%.

## Scope

Add meaningful regression tests in small, independently reviewable PRs. Start with uncovered completion logic, offline feature behavior, and statement/execution paths. Prefer existing fixtures and in-process transports. Minimal production test seams require a concrete justification.

Do not change `.octocov.yml`, coverage exclusions, generated-file filtering, the measurement command, dependencies, or unrelated production behavior. Do not inflate coverage with assertion-free calls or broad refactoring.

## Acceptance

- `make test-coverage` and `go tool cover -func=tmp/coverage.out` report at least 85% on the integrated source; verify the main CI result too.
- `.octocov.yml` continues to specify `acceptable: 80%`.
- Each PR passes `make check`, relevant focused race checks, current-head review, and all CI checks.
- Tests assert outputs, state changes, request parameters, and error contracts as appropriate; they do not introduce network/credential dependencies into offline tests.
- Preserve existing test isolation, especially global logger/configuration state.

Use `Refs` in implementation PRs until the combined result is verified. Record final coverage evidence before closing through the final PR.

apstndb:
## Candidate pool and stopping rule

Prepare more independent candidates than are needed for 85%; do not require every candidate to finish. Cursor owns prioritization, lane assignment and combined/post-merge measurement. Reconcile active ownership before splitting an existing lane. Start roughly 3-5 non-overlapping lanes when useful, with one reviewable PR per bounded change.

| Candidate | Exclusive production area and corresponding tests | Useful behavior to investigate |
| --- | --- | --- |
| A | `internal/mycli/fuzzy_finder.go` | Candidate queries, cache invalidation, variable values, selection/error handling |
| B | `internal/mycli/feature/llm/` | Offline configuration, prompts, fake-model execution, cleanup/error contracts |
| C | `internal/mycli/feature/bigquery/` and `feature/cql/` | Validation, result conversion and fake-client error behavior |
| D | `internal/mycli/execute_ddl.go` | Async results, cancellation/wait failures, buffering and operation handling |
| E | `internal/mycli/statements_explain_describe.go` | Explain/describe errors, plan export and formatting contracts |
[...]

PR #943, opened 2026-09-12
Test DDL execution and EXPLAIN/DESCRIBE paths

apstndb:
Refs #941.

Add in-process regression tests for accepted DDL operations, cancellation and error handling, and EXPLAIN/DESCRIBE results. The DDL cancellation fixture synchronizes with the fake server to avoid a scheduling-dependent deadline.

Validation: `make check` passed before publication; current-head hosted test, lint, race, vulnerability and cross-compile checks passed. TestExecuteDdlStatementsRPC passed with the race detector for 10 repetitions.

### DIFF
--- a/internal/mycli/execute_ddl_rpc_test.go
+++ b/internal/mycli/execute_ddl_rpc_test.go
@@ -1,0 +1,351 @@
+// Copyright 2026 apstndb
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package mycli
+
+import (
+	"context"
+	"errors"
+	"io"
+	"net"
+	"os"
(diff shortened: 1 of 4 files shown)


apstndb:
Reviewed 97f67162511cf3479d91fc1b07045874748040fa. Two test changes are needed before acceptance:

- [P2, source-backed] `execute_ddl_rpc_test.go:199` starts a 200ms deadline before the initial admin RPC. On a loaded runner, that deadline can expire before the operation is accepted; the function then returns the create-operation error and leaves schema generation unchanged, contrary to this test's assertions. Please synchronize cancellation with the fake observing an accepted operation/poll, using a generous outer timeout only as a deadlock guard. Avoid adding a production timing seam just to force ticker coverage.
- [P3, source-backed] At line 186, the `&&` condition allows either the wrong status code or the wrong message to pass. Assert both the expected InvalidArgument code and the injected message, since this case claims to preserve the original error.

The rest of the inspected diff is test-only and preserves the coverage policy. These findings are source-backed; no CI timing failure is claimed to have been reproduced.

apstndb:
Re-reviewed the fixing head `ac976b70fda4f62e25708080d664743ea9c89cba`. The previous review findings are addressed: DDL cancellation now waits for operation acceptance and polling before cancellation, with a timeout used only as a deadlock guard. The error-preservation case verifies both code and message.

No remaining actionable findings in the reviewed diff. TestExecuteDdlStatementsRPC passed with the race detector for 10 repetitions. Full `make check` passed on the original reviewed snapshot; the fixing head has passing hosted checks, including race, and the focused validation above. Coverage policy and production code are unchanged.

PR #944, opened 2026-09-12
Test SQL result and row processing paths

apstndb:
Refs #941.

Add regression tests for SQL result rendering, empty results, validation errors, rollback decisions, and row-processing lifecycle behavior. Assert iterator cleanup and that validation rejects invalid inputs before query execution.

Validation: `make check` passed before publication; current-head hosted test, lint, race, vulnerability and cross-compile checks passed. TestExecuteSQLImplWithQueryRunnerErrorContracts passed with the race detector for 3 repetitions.

### DIFF
--- a/internal/mycli/execute_sql_test.go
+++ b/internal/mycli/execute_sql_test.go
@@ -536,6 +536,15 @@ func TestRollbackReadWriteIfAborted(t *testing.T) {
 	}
 }
 
+func mustNotInvokeQueryRunner(t *testing.T) queryWithStatsRunner {
+	t.Helper()
+	return func(context.Context, spanner.Statement, bool, sppb.ExecuteSqlRequest_QueryMode) (*spanner.RowIterator, *spanner.ReadOnlyTransaction, error) {
+		t.Helper()
+		t.Fatal("query runner invoked; validation should have failed first")
+		return nil, nil, errors.New("unreachable")
+	}
+}
+
 func TestExecuteSQLImplWithQueryRunnerErrorContracts(t *testing.T) {
 	t.Parallel()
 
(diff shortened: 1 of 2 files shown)


apstndb:
Reviewed dc319c7de8edbb11174ddac31ae9bac91e969ac4. One test-contract gap needs tightening before acceptance:

- [P2, source-backed] In `TestExecuteSQLImplWithQueryRunnerErrorContracts`, both the `prepareFormatConfig error` and `newStatement error` cases use `runFail` and assert only a non-nil error. If the intended validation regresses and execution reaches the runner, its injected error still makes the tests pass. Use a runner that fails the test if invoked, and assert the intended validation error where practical.

Also, the `decideExecutionMode error stops iterator` case currently verifies the unsupported-mode error but does not observe iterator cleanup. Either add a meaningful cleanup assertion using an existing fixture or name the case after the behavior it actually verifies. No new production interface is needed for this.

The remaining reviewed row-processing and empty-result tests have concrete output/lifecycle assertions. No production code or coverage-policy changes are present.

apstndb:
Re-reviewed the fixing head `0e5ca777697528d219cccfab42ee5b5ed4e75f29`. The previous review findings are addressed: Validation-failure cases now reject any query-runner invocation and verify the expected error; the execution-mode failure case also observes that the returned iterator was stopped.

No remaining actionable findings in the reviewed diff. TestExecuteSQLImplWithQueryRunnerErrorContracts passed with the race detector for 3 repetitions. Full `make check` passed on the original reviewed snapshot; the fixing head has passing hosted checks, including race, and the focused validation above. Coverage policy and production code are unchanged.

apstndb:
Checked final head `b4cd74ba8fb451d4a403ce4122587b5ae0f72f45` against the accepted fixing head `0e5ca777697528d219cccfab42ee5b5ed4e75f29`. SQL execution and row-processing test files are byte-identical; incoming changes are the already-reviewed #943 and #945 tests. No actionable findings. All current-head CI checks have passed before merge.

PR #945, opened 2026-09-12
test: cover offline LLM, BigQuery, and CQL feature clients

apstndb:
Refs #941.

Add offline regression tests for LLM composition and tool use, BigQuery request/result handling, CQL type formatting, and feature/cache error paths. Use local HTTP servers and fake clients; isolate the missing-key test from ambient credentials.

Validation: `make check` passed before publication; current-head hosted test, lint, race, vulnerability and cross-compile checks passed. The missing-key test passed with the race detector for 3 repetitions while both ambient key variables contained non-secret placeholders.

### DIFF
--- a/internal/mycli/feature/bigquery/client_offline_test.go
+++ b/internal/mycli/feature/bigquery/client_offline_test.go
@@ -1,0 +1,218 @@
+// Copyright 2026 apstndb
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//	http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package bigquery
+
+import (
+	"context"
+	"encoding/json"
+	"io"
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"testing"
+
+	bq "cloud.google.com/go/bigquery"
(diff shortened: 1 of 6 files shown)


apstndb:
Reviewed 2607722b7f42454f8dc233532e5bedc80999c8d0. One environment-isolation issue needs fixing before acceptance:

- [P2, source-backed] The `missing API key` subtest in `compose_offline_test.go` passes an empty `ClientConfig.APIKey` without clearing `GOOGLE_API_KEY` and `GEMINI_API_KEY`. The installed GenAI SDK documents that it reads those environment variables as fallbacks. On a developer machine with either key set, this does not exercise the missing-key rejection and instead enters the normal compose path, so the offline test can fail for an unrelated reason. Move this case out of the parallel parent, clear both variables with `t.Setenv`, and retain a local/failing HTTP transport guard if useful. No real credentials or network request are needed to verify the fix.

The other inspected changes reuse local HTTP/fake-client fixtures and preserve production code, dependencies and the 80% CI floor.

apstndb:
Re-reviewed the fixing head `1998a28e169c29e4530f9f51420f532cdcce4bcb`. The previous review findings are addressed: The missing-key case is a sequential top-level test that clears both SDK fallback variables and uses a failing HTTP transport. It now consistently exercises the missing-key rejection.

No remaining actionable findings in the reviewed diff. The missing-key test passed with the race detector for 3 repetitions while both ambient key variables contained non-secret placeholders. Full `make check` passed on the original reviewed snapshot; the fixing head has passing hosted checks, including race, and the focused validation above. Coverage policy and production code are unchanged.

apstndb:
Checked final head `5a409edfef1750d3a73ecd8d389b66a2e26a3657` after the additive main merge. This PR's feature tests are byte-identical to the accepted fixing head `1998a28e169c29e4530f9f51420f532cdcce4bcb`; the only incoming changes are the already-reviewed #943 tests. No actionable findings. All current-head CI checks have passed.

PR #951, opened 2026-09-12
Pass operation output explicitly through statement execution

### LINKED ISSUE #911: Track bounded simplification of parsing, output, sessions and transaction ownership
## Goal

Make the existing implementation simpler through bounded, independently reviewable PRs. This tracks the remaining September 2026 architecture proposals, concretized against main commit `96a8a60ab5df71447b30db1826d4da2d6677c42b`. Each child specifies a chosen implementation, deletion target, behavior boundaries and validation. It is a work sequence, not permission for a broad rewrite.

## Already complete

- #908: completion cache map.
- #909: shared DUMP traversal and explicit internal writer.
- #910: transaction-options builder removal.

These are completed baselines and must not be implemented again. Existing #380 is reused for source preservation. #790 remains the separate BIGQUERY typed-result contract; #738 and #538 are completed background work, not reopened tasks.

## Independently startable work

- [x] #380 — Preserve original SQL fragments for SOURCE and batch execution
- [x] #913 — Pass narrow export options to the shared row writer and remove typed replay settings copies
- [x] #914 — Make query-cache publication explicit and independent of SQL table-name detection
- [x] #916 — Stop and join progress output before the result sink emits its first bytes
- [x] #917 — Represent each Result body with one closed variant while retaining shared metadata
- [x] #919 — Give each Session a fixed connection identity independent of live registry settings
- [x] #921 — Replace the single-use TransitTransaction wrapper with a direct pending transition
- [x] #922 — Verify delayed heartbeat ticks cannot act on a replacement transaction owner

## Work with prerequisites

- [x] #912 — Share native statement-kind mapping between semantic and lexical parsing
- [x] #915 — Replace per-query systemVariables copies with narrow rendering overrides
- [x] #918 — Pass operation output explicitly through statement execution and remove Session output mutation
- [x] #920 — Validate replacement sessions before publishing live connection values and callbacks
- [x] #923 — Preserve the logical transaction context across pending activation and own SET LOCAL undo there
- [x] #924 — Store automatic DML on its transaction context and remove parallel generation bookkeeping

## Sequence and implementation boundaries

- Export-writer inputs + explicit query-cache publication → query rendering overrides → explicit operation output. Progress stop/join also precedes explicit operation output.
- Fixed session construction identity → candidate validation and adoption.
- Direct pending transition + completed heartbeat investigation → stable logical transaction lifetime → automatic-DML queue ownership.
- Source-envelope changes land before parser-mapping cleanup. Result-body migration lands before the broad explicit-output signature migration. Both orders are recorded as blocked-by links so implementations do not overlap in the shared files; these are sequencing prerequisites, not semantic contract dependencies.
- Use separate worktrees per PR. Parallelize independent preparation; serialize heavy gates and integrate changes against refreshed main. Keep each issue's old path removal in the same PR as its replacement.

The heartbeat item is an **unreproduced scheduling hypothesis** with a deterministic investigation plan. It is not a confirmed bug or a reason to rewrite locking. The query-cache item identifies a source-backed behavior discrepancy and requires a red/green regression before its fix. Other items are structural refactors except the explicitly documented SOURCE/batch echo and pre-output progress improvements.

## Completion criteria

- [x] Every child has a merged PR satisfying its own behavior and deletion criteria, or a documented evidence-based disposition for the investigation.
- [x] No temporary parallel representations, obsolete settings copies or ownership shims remain from this series.
- [x] Existing output, transaction, SET LOCAL, feature, DUMP and MCP contracts remain covered; intentional behavior changes are isolated and explained.
- [x] Required checks and independent current-head review are recorded per PR. No performance or correctness claims are inferred from aggregate architecture review.

No package extraction campaign, single-implementation interfaces, provider-neutral result framework, dependency upgrades, release or external-repository issue creation is included.

### LINKED ISSUE #918: Pass operation output explicitly through statement execution and remove Session output mutation
## Problem and outcome

The current scoped Session.output save/restore is safe under serialized execution, but makes nested output ownership implicit. The operation destination and lazy width resolver should travel with the call instead of mutating a long-lived Session. This is a structural simplification, not a claim of a reproduced data race.

## Starting points

Inspected on main commit `96a8a60ab5df71447b30db1826d4da2d6677c42b` after #908, #909 and #910. Start the implementation from refreshed main and reconcile moved code; the permalinks below are a verified baseline, not a requirement to branch from this old commit.

- [internal/mycli/output_context.go:33](https://github.com/apstndb/spanner-mycli/blob/96a8a60ab5df71447b30db1826d4da2d6677c42b/internal/mycli/output_context.go#L33)
- [internal/mycli/statement_processing.go:41](https://github.com/apstndb/spanner-mycli/blob/96a8a60ab5df71447b30db1826d4da2d6677c42b/internal/mycli/statement_processing.go#L41)
- [internal/mycli/cli.go:516](https://github.com/apstndb/spanner-mycli/blob/96a8a60ab5df71447b30db1826d4da2d6677c42b/internal/mycli/cli.go#L516)
- [internal/mycli/execute_sql.go:159](https://github.com/apstndb/spanner-mycli/blob/96a8a60ab5df71447b30db1826d4da2d6677c42b/internal/mycli/execute_sql.go#L159)
- [internal/mycli/statements_dump.go:109](https://github.com/apstndb/spanner-mycli/blob/96a8a60ab5df71447b30db1826d4da2d6677c42b/internal/mycli/statements_dump.go#L109)

## Implementation plan (one PR)

1. Promote the small outputContext value to an exported OperationOutput type (so feature statement signatures can name it), with writer and lazy screen-width resolver. Resolve fallback writer/width at the execution entry point; keep the resolution helper concrete and small.
2. Change the existing Statement.Execute signature to Execute(context.Context, *Session, OperationOutput). Update core and the three feature implementations in the same PR. A statement that does not emit output ignores the argument; do not add parallel output-aware interfaces or adapters.
3. Carry the same value through Session/SessionHandler execution, RunBatchStatement re-entry, queryExecution, partitioned queries, struct-row streaming, DUMP and shell meta-command output. Nested execution must explicitly forward the caller's operation value.
4. Keep Session.ExecuteStatement as a convenience entry that resolves the default output; its explicit-output entry routes directly without withOutput. Preserve USE/DETACH handler dispatch. For DUMP internal buffering, override a local operation copy's writer while retaining the outer publication destination and width behavior.
5. Remove Session.output, withOutput and hidden outputWriter/displayWidthFor reads. Keep output/width lazy at the same point and resolve terminal width against the original destination, not the pager pipe. Preserve direct-call nil-output buffered fallback.
6. Result bodies must not retain output writers. Continue using the existing resultSink and MCP capped writer without altering their io.Writer contracts.

## Acceptance criteria

- [ ] Explicit caller output wins over StreamManager for every streaming route, partitioned queries, DUMP, virtual rows and nested RUN BATCH.
- [ ] MCP receives all captured bytes without leaked SQL on JSON-RPC stdout; its capped writer still reports full writes when discarding bytes beyond its limit.
- [ ] Tee/redirection, pager/fence ordering, late width resolution, meta-command output and success/error prefix behavior stay unchanged.
- [ ] Nested output overrides do not affect the next statement; no session field is saved/replaced/restored for output.
- [ ] Search confirms every nested Execute call forwards OperationOutput, all feature builds compile, and no new dispatch interface or context.Value output carrier exists.

## Scope and dependencies

Prerequisites: #915, #916, #917 The body prerequisite serializes broad producer migrations; it does not mean the output contract depends semantically on the body representation.

No concurrent statement support, Result body redesign, new output policy, feature compatibility layer or changes to transaction commit/output timing.

Tracking: #911. Related issues are context unless explicitly listed as prerequisites.

## Validation and handoff

Read the repository AGENTS.md and relevant existing tests before editing. Use a separate feature worktree. Extend behavior tests at the listed seams; do not add tests that merely mirror helpers or weaken existing assertions to accommodate the refactor.

```sh
go test -short ./internal/mycli/... -run 'Output|ResultSink|Dump|Partitioned|StructRows|ExecuteStatement'
make check-race
make check
```

Capture each command's full output to a file and check its exit code; do not infer success from a tail/grep pipeline. Run the focused checks during development and the full required gate once the change is complete, using the Go/linter versions and emulator setup specified by current CI. Record changed files, deleted paths/abstractions, test commands/results and any deviations in the PR. No dependency upgrades or unrelated cleanup. One issue is one independently mergeable PR; dependencies must land first. Required CI and independent current-head review still apply before merge.

apstndb:
Statement execution now carries a concrete `OperationOutput` through the existing `Execute` call chain. This removes `Session.output` and its save/restore wrappers while preserving caller-provided writers, lazy width resolution, direct-call defaults, DUMP buffering and MCP output isolation. Output-free abort rollback calls the transaction manager directly.

Regression tests exercise actual nested RUN BATCH success/error paths, the next statement's destination and buffered DUMP isolation. An intentional dropped forwarding argument is detected by the new test. Independent focused tests and `make check` passed at 71eaa172d7e863a8c732d0e6517b9f05be43aaea, along with current-head review and required CI (including `go test -short -race ./...`). The CI coverage floor remains 80%.

This completes the final child of the bounded simplification series in #911.

Fixes #918.
Fixes #911.

### DIFF
--- a/internal/mycli/cli.go
+++ b/internal/mycli/cli.go
@@ -596,7 +596,7 @@ func (c *Cli) executeStatement(ctx context.Context, stmt Statement, interactive 
 		defer sink.abort()
 		outW = sink
 	}
-	out := outputContext{
+	out := OperationOutput{
 		w: outW,
 		// Resolve the width against the caller's writer, not the sink: the
 		// pager pipe is never a terminal.
--- a/internal/mycli/cli_input_test.go
+++ b/internal/mycli/cli_input_test.go
@@ -342,7 +342,7 @@ type abortStatement struct{}
 
 func (abortStatement) isDetachedCompatible() {}
 
-func (abortStatement) Execute(context.Context, *Session) (*Result, error) {
+func (abortStatement) Execute(context.Context, *Session, OperationOutput) (*Result, error) {
 	return nil, spanner.ToSpannerError(status.Error(codes.Aborted, "txn aborted"))
 }
 
@@ -430,7 +430,7 @@ type timestampResultStatement struct{ ts time.Time }
 
 func (timestampResultStatement) isDetachedCompatible() {}
 
-func (s timestampResultStatement) Execute(context.Context, *Session) (*Result, error) {
+func (s timestampResultStatement) Execute(context.Context, *Session, OperationOutput) (*Result, error) {
 	return &Result{
 		ReadTimestamp:   s.ts,
 		CommitTimestamp: s.ts,
@@ -542,7 +542,7 @@ type keepVariablesStatement struct{}
 
 func (keepVariablesStatement) isDetachedCompatible() {}
 
-func (keepVariablesStatement) Execute(context.Context, *Session) (*Result, error) {
+func (keepVariablesStatement) Execute(context.Context, *Session, OperationOutput) (*Result, error) {
 	return &Result{
 		KeepVariables: true,
 		ReadTimestamp: time.Date(2026, 9, 12, 0, 0, 0, 0, time.UTC),
--- a/internal/mycli/cli_mcp_cancellation_test.go
+++ b/internal/mycli/cli_mcp_cancellation_test.go
@@ -33,7 +33,7 @@ type mcpAdmissionBlockingStatement struct {
 	release <-chan struct{}
 }
 
-func (s *mcpAdmissionBlockingStatement) Execute(ctx context.Context, _ *Session) (*Result, error) {
+func (s *mcpAdmissionBlockingStatement) Execute(ctx context.Context, _ *Session, out OperationOutput) (*Result, error) {
 	close(s.entered)
 	select {
 	case <-s.release:
--- a/internal/mycli/cli_output_error_test.go
+++ b/internal/mycli/cli_output_error_test.go
@@ -149,8 +149,8 @@ type resultErrorStatement struct{ cause error }
 
 func (resultErrorStatement) isDetachedCompatible() {}
 
-func (s resultErrorStatement) Execute(_ context.Context, session *Session) (*Result, error) {
-	if _, err := io.WriteString(session.outputWriter(), "BODY\n"); err != nil {
+func (s resultErrorStatement) Execute(_ context.Context, _ *Session, out OperationOutput) (*Result, error) {
+	if _, err := io.WriteString(out.Writer(), "BODY\n"); err != nil {
 		return nil, err
 	}
 	return nil, s.cause
--- a/internal/mycli/cli_progress_test.go
+++ b/internal/mycli/cli_progress_test.go
@@ -495,8 +495,8 @@ type streamThenErrorStatement struct{}
 
 func (streamThenErrorStatement) isDetachedCompatible() {}
 
-func (streamThenErrorStatement) Execute(_ context.Context, session *Session) (*Result, error) {
-	w := session.outputWriter()
+func (streamThenErrorStatement) Execute(_ context.Context, _ *Session, out OperationOutput) (*Result, error) {
+	w := out.Writer()
 	if w == nil {
 		return nil, errors.New("no output writer")
 	}
--- a/internal/mycli/directed_read_acceptance_test.go
+++ b/internal/mycli/directed_read_acceptance_test.go
@@ -226,7 +226,7 @@ func TestDirectedReadSelectedROPathsABClear(t *testing.T) {
 			}
 			srv.takeRequests()
 			srv.takePartitionQueries()
-			result, err := runPartitionedQuery(ctx, session, "SELECT 'observed'")
+			result, err := runPartitionedQuery(ctx, session, "SELECT 'observed'", session.resolveOperationOutput(OperationOutput{}))
 			if err != nil || result == nil || result.AffectedRows != 1 || result.PartitionCount != 1 {
 				t.Fatalf("%s partition result=%v error=%v", sel.name, result, err)
 			}
--- a/internal/mycli/directed_read_matrix_test.go
+++ b/internal/mycli/directed_read_matrix_test.go
@@ -553,7 +553,7 @@ func TestDirectedReadDumpSnapshotWire(t *testing.T) {
 					}
 					srv.takeRequests()
 					srv.takeBegins()
-					result, err := executeDump(ctx, session, dumpModeTables, []tableID{{Name: "T"}})
+					result, err := executeDump(ctx, session, dumpModeTables, []tableID{{Name: "T"}}, OperationOutput{w: writer})
 					if mode == "cyclic reject" {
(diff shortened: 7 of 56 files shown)


apstndb:
Review of HEAD 19dbfb62a4488f723060b15927b9dd0d17cd8454 found one acceptance-test gap to address before merging.

**Exercise actual nested output dispatch.** `internal/mycli/output_context_test.go:112` says it verifies nested execution, but only copies `OperationOutput` and compares writer pointers. It never executes a statement, RUN BATCH or DUMP, and does not check the next statement. This does not cover the new forwarding boundary required by #918.

I verified this using a Go overlay: replacing the nested call at `internal/mycli/statements.go:727` with `session.ExecuteStatement(ctx, batch)` still passes `TestOperationOutput|TestExecuteStatementWithOutput|TestMCPHandler_streamed`. A small probe using a client-less database-mode session, `batch.SetCurrent(&ShowVariablesStatement{})`, and `ExecuteStatementWithOutput(..., &RunBatchStatement{}, callerOutput)` passes on the actual source, but fails with that mutation: 0 bytes reach the caller and 2106 bytes reach the global stream. The probe then executes an ordinary next statement to verify its default destination. This is a reproduced test gap, not an observed bug in the submitted forwarding implementation.

Replace the copy-only test with behavior coverage through the actual nested execution entry (including success/error and a subsequent statement); retain or extend the existing buffered-DUMP isolation coverage. Verify that intentionally dropping the RUN BATCH output value makes the new regression test fail. Keep test synchronization deterministic.

One small source-contract cleanup: PROFILE's two calls to `rollbackReadWriteIfAborted` at `statements_explain_describe.go:455,469` manufacture `OperationOutput{}`. Rollback emits no statement output. Inside the helper's existing live-RW/Aborted guard, call `session.txn.RollbackReadWriteTransaction(ctx)` directly and remove the artificial output parameter from this helper and its callers. Preserve the original abort and rollback error joining; the existing rollback tests should still apply. This keeps output forwarding complete at the remaining actual nested Statement.Execute calls without plumbing output through buffered EXPLAIN helpers.

Please retain the current branch/history, run the focused tests and required gates on the fixing HEAD, then handle its CI. No broader refactor is requested.

apstndb:
The review requests are addressed in 71eaa172d7e863a8c732d0e6517b9f05be43aaea. Independent re-review of this exact HEAD: no actionable findings.

The replacement tests exercise actual RUN BATCH dispatch, a nested error with its expected cause, the next statement's default destination, and buffered DUMP isolation. Independent focused tests pass. Replacing the nested forwarding call with `ExecuteStatement(ctx, batch)` now makes the new regression test fail for the expected caller/global destination mismatch. Rollback directly uses the existing transaction manager and preserves error joining, removing the artificial output argument.

Independent `make check` passed with Go 1.26.8, golangci-lint 2.13.2 and the Colima-backed emulator, including all feature packages. Required CI checks passed on this HEAD; no unresolved review threads or pending reviews remain. Reviewed the complete output migration, explicit writer priority, lazy width, DUMP, shell/MCP and nested dispatch paths. The existing serialized execution contract is unchanged.

PR #952, opened 2026-09-13
Add private SAVEPOINT journal and typed fingerprints

apstndb:
Refs #404.

Private SAVEPOINT journal and typed fingerprint helpers only. No `CLI_SAVEPOINT_SUPPORT`, no client-side statements, and no transaction-manager integration.

This PR introduces:
- Frozen SQL/parameters/query options and mutation writes (cloned protobuf graphs, not session aliases)
- SHA-256 fingerprints over field names, complete recursive types, ordered rows, EOS, affected counts, and Batch DML count vectors
- 16 MiB journal byte accounting that rejects new payload without invalidating existing markers

Prerequisite: none (targets `main` at `3f2d97bbab8d77f9096cb898563079464dcd5a6a`). Public feature stays inaccessible until PR5.

<div><a href="https://cursor.com/agents/bc-e4aa8f59-1f19-52a9-8598-f707b0c118e3?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-e4aa8f59-1f19-52a9-8598-f707b0c118e3&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

Issue #404, opened 2025-07-19
feat: Support SAVEPOINT, RELEASE SAVEPOINT, and ROLLBACK TO SAVEPOINT


apstndb:
This feature is about implementing transaction savepoints.

**Reference (`java-spanner`):**
`java-spanner` supports the standard SQL commands `SAVEPOINT <name>`, `RELEASE SAVEPOINT <name>`, and `ROLLBACK TO SAVEPOINT <name>`. This allows for fine-grained control over transaction rollbacks.

**Implementation suggestion for `spanner-mycli`:**
This would require significant changes to the internal transaction management logic to track and manage multiple rollback points within a single transaction.

PR #953, opened 2026-09-13
Add SAVEPOINT operation receipts on existing row paths

apstndb:
Refs #404.

Prerequisite: PR #952 (`cursor/savepoint-01-journal`).

Optional operation receipts on the existing buffered and streaming row-consumption paths. Raw rows are observed once before transform, and terminal status is recorded even on zero rows or error. A nil observer is a no-op, so current callers are unchanged.

This PR does not:
- Add `CLI_SAVEPOINT_SUPPORT` or SAVEPOINT statements
- Attach a journal to `transactionContext`
- Introduce a manager-global receipt field

Public SAVEPOINT remains inaccessible until PR5.


apstndb:
Reviewed HEAD `493bbe768ddc4a1834f17640523337b7dd1cd65a` against prerequisite `020ce5258fa56851505efee26f872ff7ab4d6160`. Two issues need addressing before accepting this receipt layer:

1. **Keep DML completion distinct from query completion.** [consumeRowIterObserving](https://github.com/apstndb/spanner-mycli/blob/493bbe768ddc4a1834f17640523337b7dd1cd65a/internal/mycli/row_iter.go#L66) unconditionally finalizes the query fingerprint. A later `FinishDML(affected, nil)` then hits [the finalized fast path](https://github.com/apstndb/spanner-mycli/blob/493bbe768ddc4a1834f17640523337b7dd1cd65a/internal/mycli/savepoint_receipt.go#L116), returning that old hash without including the affected count. An independent fake-gRPC/Go-overlay probe runs the actual observing collector twice over identical returned rows and completes with counts 1 and 2: both fingerprints are equal. The direct `FinishDML` unit test passes because it bypasses the collector. Adjust the completion contract so the intended DML consumer cannot silently drop counts, and test through that consumer. This is a **runtime-reproduced private-layer defect**, currently hidden because transaction wiring is still disabled.

2. **Test the actual buffered and two streaming forwarding paths.** [TestSavepointReceiptBufferedAndStreamingOnce](https://github.com/apstndb/spanner-mycli/blob/493bbe768ddc4a1834f17640523337b7dd1cd65a/internal/mycli/savepoint_receipt_test.go#L82) calls the underlying helpers directly; it does not exercise the new `queryExecution.Receipt` forwarding in `executeWithBuffering`, `runSpanvalueRowIterator`, or `runSpanvalueRowIteratorWithProcessor`. Removing both `withRowIteratorReceipt(qe.Receipt)` arguments and replacing the buffered receipt with nil through a Go overlay still passes all tests selected by `Savepoint|Replay|OperationReceipt`. Add tests entering those actual wrappers, with real fake-gRPC iteration, including zero rows and a writer/transform error or cancellation. They should fail when any forwarding argument is removed. **Runtime-reproduced validation gap.**

Focused tests on the real source passed independently. Please also carry the prerequisite #952 fixes forward additively. Keep public SAVEPOINT inaccessible while the stack is incomplete.

apstndb:
Verified the fixes in `914df5897f142eafa3daece4b399676c56bcee1d`, including the additive prerequisite merge of #952 at `4781311e141de7564721586e93c10404eec309db`. Re-reviewed the complete receipt-layer diff against that prerequisite.

Both earlier findings are addressed: collection no longer prematurely finalizes DML as a query, and tests now enter the actual buffering, streaming writer and streaming processor wrappers, including empty/error cases.

Independent validation on this exact HEAD:
- The original collector/DML affected-count probe passes.
- Removing receipt forwarding now fails the committed new path test, as intended.
- Focused tests and `make check` passed on macOS with Go 1.26.8, golangci-lint 2.13.2 and Colima-backed integration tests.
- Current GitHub CI checks passed; no unresolved review threads were found.

**No further actionable findings in this PR's scope.** Transaction admission, journal ownership, reconstruction, and public SAVEPOINT behavior remain subject to review in the following PRs. No merge was performed.

PR #954, opened 2026-09-13
Record explicit-transaction operations in the SAVEPOINT journal

apstndb:
Refs #404.

Prerequisite: PR #953 (`cursor/savepoint-02-receipts`).

Owner integration for the private SAVEPOINT journal on explicit transactions:

- Attach the journal at explicit `BEGIN` (pending, RW, RO). Implicit one-statement RW does not get a journal.
- Freeze the RW constructor options (isolation, priority, tag, commit options, read-lock, exclude from change streams).
- Record SQL queries, ordinary DML/`THEN RETURN`, automatic and manual Batch DML, and `MUTATE` through the existing receipt paths. Heartbeat and PLAN-mode EXPLAIN are not recorded.
- Reserve journal payload before RPC or enqueue. Queued automatic DML is not a replay entry until BatchUpdate succeeds.

Additive review follow-up: bind admission/completion to an owner/attempt capture token, reserve known Batch DML and MUTATE payload before effects, and parse MUTATE once including `KEY_RANGE`. Local journal-full rejection keeps the explicit owner.

Additive follow-up: the shared DML helper accepts the caller's effective QueryMode. Ordinary DML respects `CLI_QUERY_MODE`; EXPLAIN ANALYZE DML forces PROFILE. That option is frozen for replay and wire-asserted with capture enabled and disabled.

Additive follow-up: automatic-queue reservation treats an empty queue as zero reserved bytes and uses that same amount on enqueue, flush, and discard. Lifecycle tests require retained bytes to match marker/journal/queue payload after each transition, including an exact-limit first enqueue and repeated discard.

Public `CLI_SAVEPOINT_SUPPORT` and SAVEPOINT statements remain unregistered. Physical replacement and recovery-required are later stack PRs.


apstndb:
Reviewed HEAD `9e9ecd2756e1c445e28fb07fd9ebb64d6fa6fe0b` against prerequisite `914df5897f142eafa3daece4b399676c56bcee1d`. Three issues need addressing before accepting owner integration. All three reproduce with independent Go-overlay tests using the existing fake-gRPC heartbeat harness; the committed focused tests pass.

1. **Bind admission, observation and completion to the exact operation and owner/attempt.** [The run-error path](https://github.com/apstndb/spanner-mycli/blob/9e9ecd2756e1c445e28fb07fd9ebb64d6fa6fe0b/internal/mycli/execute_sql.go#L339) calls `finishQueryCapture(err)` even if that query was never admitted. Start query A with `RunQueryWithStats` and leave its iterator in flight, then execute query B through `executeSQLImplWithVars`. B correctly reports `a query is already in flight`, but its cleanup erases A's pending capture, changes inFlight from 1 to 0, and releases all 48 reserved bytes. A can then finish without a replay entry. Completion currently fetches whatever `tm.tc.pending` happens to contain, so it also cannot distinguish an old owner/physical attempt from a replacement. Return/carry an operation token alongside the iterator/receipt and validate its exact owner/attempt on both success and error. Rejected or uncaptured operations must never complete another operation's capture. Add deterministic real-wrapper tests, including stale completion after owner/attempt replacement. **Runtime-reproduced correctness defect.**

2. **Reserve known payload before batch RPCs or mutation buffering, preserving the owner on local admission failure.** [executeBatchDML](https://github.com/apstndb/spanner-mycli/blob/9e9ecd2756e1c445e28fb07fd9ebb64d6fa6fe0b/internal/mycli/execute_dml.go#L129) sends BatchUpdate before freezing/reserving; MUTATE similarly reserves only after BufferWrite. With an explicit capture-enabled transaction's retainedBytes at the limit, one actual ExecuteBatchDml RPC is recorded before errSavepointJournalFull, and the owner is retired. This violates the accepted pre-RPC admission contract. Automatic enqueue also reserves statement payload without the known fingerprint/count-vector overhead, leaving a predictable failure until after flush. Freeze/reserve all known bytes before effects, distinguish local rejection from an executed-operation failure, and retain the existing journal/markers/owner on rejection. Test RPC counts and owner identity, not just the error text. **Runtime-reproduced batch defect; mutation/automatic-queue ordering is source-backed.**

3. **Preserve existing MUTATE key-range support and parse inputs once.** [freezeDeleteMutate](https://github.com/apstndb/spanner-mycli/blob/9e9ecd2756e1c445e28fb07fd9ebb64d6fa6fe0b/internal/mycli/savepoint_capture.go#L317) rejects every CallExpr. `MUTATE T DELETE KEY_RANGE(start_closed=>(1), end_open=>(10))` succeeds through the existing parser but fails under owner capture with `key range is not journaled`, aborting the transaction. The accepted design explicitly includes resolved keys/ranges. Also, Execute first calls parseMutation and then reparses via freezeMutate. Refactor the existing parsing path to produce frozen resolved inputs and SDK mutations together, including range bounds/kind, so capture preserves supported semantics without a duplicate parser. Add typed-key/range cases and immutability/wire validation. **Runtime-reproduced compatibility defect; duplicate parsing is source-backed.**

Keep this layer private. Apply fixes additively on #954 and carry them forward before accepting reconstruction/public layers. The existing worker should continue the authorized stack at a safe checkpoint; no duplicate lane or bot review is needed.

apstndb:
Additional independent finding at the same HEAD `9e9ecd2756e1c445e28fb07fd9ebb64d6fa6fe0b`, also present in #955: **preserve forced PROFILE for EXPLAIN ANALYZE DML even when CLI_QUERY_MODE=WITH_STATS.**

[executeExplainAnalyzeDML now calls runUpdateOnTransaction](https://github.com/apstndb/spanner-mycli/blob/9e9ecd2756e1c445e28fb07fd9ebb64d6fa6fe0b/internal/mycli/statements_explain_describe.go#L597), which uses the ordinary query-mode setting. The previous path explicitly forced PROFILE so the query plan is available. An independent fake-gRPC probe entering executeExplainAnalyzeDML with WITH_STATS observes WITH_STATS on the wire instead of PROFILE, both with savepoint capture enabled and **with it disabled**. On a real server this mode omits the requested plan and the command cannot return its explanation. This is an existing-feature regression, not merely an incomplete private savepoint path.

Allow the shared DML execution/capture helper to receive the caller's effective mode (ordinary DML respects CLI_QUERY_MODE, EXPLAIN ANALYZE forces PROFILE), freeze that actual option for replay, and add a wire assertion covering both support settings. **Runtime-reproduced request regression; real-server missing-plan consequence is source/contract-backed.**

apstndb:
## Execution receipt (not a review or acceptance)

Assigned SAVEPOINT lane is still running; this is not a new assignment.

- Stage 3 [#954](https://github.com/apstndb/spanner-mycli/pull/954) branch `cursor/savepoint-03-owner-journal` published HEAD `2e84f30a1a59d1d6cc9250b7cff11bff87ebccb8`.
- Stage 4 [#955](https://github.com/apstndb/spanner-mycli/pull/955) branch `cursor/savepoint-04-reconstruction` published HEAD still `1027cb60ba6ef9fc85c1c5ee1dbfc141f76a7708`; reconstruction follow-up is in progress and not yet pushed.
- Unpublished public-layer work remains local and is not being published yet.

Review comments received (bodies not repeated):
- https://github.com/apstndb/spanner-mycli/pull/954#issuecomment-5652399205 — first owner-journal review; published in `2e84f30`.
- https://github.com/apstndb/spanner-mycli/pull/955#issuecomment-5652418139 — reconstruction review; current work.
- https://github.com/apstndb/spanner-mycli/pull/954#issuecomment-5652431380 — EXPLAIN ANALYZE query-mode regression; queued after the current reconstruction work; not in `2e84f30`.

Next: finish the #955 additive fixes, then the remaining #954 query-mode regression, then the public feature only after these blockers and the accepted end-to-end matrix. No blocking design question. No merge.

apstndb:
Re-reviewed `2e84f30a1a59d1d6cc9250b7cff11bff87ebccb8`. The original three independent probes now pass: rejected query B preserves A's capture, full manual-batch budget rejects before RPC and retains its owner, and the existing DELETE range is accepted. Thank you for the execution receipt; the separately reported PROFILE regression remains queued and is not treated as fixed here.

One accounting defect remains in the new automatic-queue reservation: [enqueueFrozenAutomaticDMLLocked](https://github.com/apstndb/spanner-mycli/blob/2e84f30a1a59d1d6cc9250b7cff11bff87ebccb8/internal/mycli/savepoint_capture.go#L361) subtracts `batchReplayAccounted(nil)`, which is **40**, although an empty queue has no reservation. The first enqueue therefore omits the fixed entry/fingerprint overhead. Meanwhile [dropQueued](https://github.com/apstndb/spanner-mycli/blob/2e84f30a1a59d1d6cc9250b7cff11bff87ebccb8/internal/mycli/savepoint_journal.go#L143) still releases only statement payload and leaks the newly reserved count bytes.

Independent probe: reserve marker `keep` (12 bytes), enqueue `UPDATE T SET v=2 WHERE id=1`; retainedBytes is 47 but the marker plus complete future batch entry is 87. DiscardAutomaticDML leaves 20 instead of the original 12. Flush likewise records a payloadBytes amount larger than what was actually reserved. Repeated enqueue/discard consumes budget with no retained queued data; completed batches undercount their actual retained entries.

Treat the empty queue's reserved amount as zero and use the same exact accounting on enqueue, successful flush, failed flush and discard. Prefer one explicit reservation amount or a consistent nonempty-queue helper; do not hide the mismatch with the underflow clamp. Add lifecycle assertions that reservation equals marker/journal/queue payload after each transition, including an exact-limit first enqueue and repeated discard. **Runtime-reproduced accounting defect.** Keep the existing worker and stack; no duplicate implementation lane is needed.

apstndb:
Re-reviewed exact HEAD `98c331b1a6865ca770586b16247ed9c4a3cf3f4b`. No further actionable findings in this private owner-journal layer.

The capture-ownership, pre-RPC admission and DELETE key-range findings are fixed by `2e84f30`; forced EXPLAIN ANALYZE PROFILE is fixed by `99800ab`; automatic-queue reservation conservation is fixed by `98c331b`. Independent regression probes pass, including enabled/disabled PROFILE wire mode and first enqueue/discard accounting. The new committed lifecycle tests cover empty, exact-limit, repeated discard, failed completion and successful flush cases.

Independent `make check` passed on this exact head (Go 1.26.8, macOS, Colima/emulator); all substantive current-head CI checks passed, and there are no unresolved inline review threads. This accepts this PR's private integration scope. Reconstruction/recovery and the public feature remain under separate review in #955/#956; this is not acceptance of those layers or a merge action.

PR #955, opened 2026-09-13
Replace the physical RW attempt on ROLLBACK TO SAVEPOINT

apstndb:
Refs #404.

Prerequisite: PR #954 (`cursor/savepoint-03-owner-journal`).

Physical replacement and recovery-required for the private SAVEPOINT journal:

- `ROLLBACK TO` keeps the logical owner, rolls back the old RW handle with a bounded cleanup context, reconstructs the verified prefix on a new SDK transaction using frozen constructor options, then keeps the target marker and drops later markers including equal-position names.
- Replay fingerprints the newly observed DML affected-row count and returned rows. A mismatch ends the logical transaction and does not commit the candidate.
- Heartbeat is bound to the originating physical attempt. `EnableHeartbeat` captures that attempt before launching the goroutine; the RPC-acquisition lock also refuses a cancelled originating context. A delayed tick or delayed startup after replacement emits no RPC; the restarted heartbeat runs on the new attempt.
- Reconstruction recovery is restricted to explicit RW. A failed RO query keeps the snapshot handle and does not set recovery-required.
- `SAVEPOINT` reserves marker bytes before flushing automatic DML. The reservation owner is the local `replayState`, so a first-marker automatic DML flush that retires `tm.tc` still releases the reservation without panicking. Local journal-full rejection leaves the queue and owner intact.
- A statement failure after a completed marker discards the physical RW attempt, retains the logical owner and earlier journal, and sets recovery-required. `COMMIT`, new savepoints, and new SQL are rejected until `ROLLBACK TO`.
- Pending and read-only owners truncate markers only.

Owner/attempt binding covers receipt completion and post-operation failure handling. `HandleOwnerFailure` applies the captured owner/attempt check under the same lock as recovery or abort rollback. Unadmitted and single-use operations cannot poison an explicit owner. Capture-off abort cleanup remains the nil-token `rollbackOnAbort` path.

EXPLAIN ANALYZE SELECT, including `CLI_QUERY_MODE=PROFILE`, uses that same owner/attempt-bound completion and failure contract. A failed stream after a completed marker enters recovery-required instead of fully retiring the owner; `ROLLBACK TO` restores the earlier prefix.

PLAN-mode EXPLAIN SELECT and DESCRIBE SELECT carry the same owner/attempt identity and in-flight lifetime through collection and error, so a failed plan after a completed marker also enters recovery-required. Successful PLAN operations are not reserved or committed into the replay journal: they are not hashed and they do not manufacture a data replay entry. Single-use and read-only isolation is unchanged.

Negative fake-gRPC coverage: changed replayed DML counts (1→2 and 1→0), changed returned rows, equal-position marker truncation on pending/RO/RW, delayed heartbeat tick and delayed heartbeat startup after `ROLLBACK TO`, RO query failure without recovery, marker reservation before automatic flush, first-marker automatic DML flush that retires the owner without panic, later-marker flush that enters recovery without recording the new marker, cancelled-command reconstruction cleanup of the candidate after it has an id and without candidate commit, a retired iterator from owner A consumed against live owner B, single-use failure against an explicit owner with a marker, EXPLAIN ANALYZE / `CLI_QUERY_MODE=PROFILE` stream failure (`PermissionDenied` and `ABORTED`) after a completed marker, and EXPLAIN/DESCRIBE SELECT PLAN failure (`PermissionDenied` and `ABORTED`) after a completed marker with successful `ROLLBACK TO`. Successful PLAN after a marker does not grow the journal.

Carries the #954 QueryMode follow-up: ordinary DML respects `CLI_QUERY_MODE`; EXPLAIN ANALYZE DML forces PROFILE on the wire and freezes that option for replay, with capture enabled and disabled.

Carries the #954 automatic-queue reservation follow-up: an empty queue is zero reserved bytes; enqueue, flush, and discard use the same nonempty-queue amount. Lifecycle tests require retained bytes to match marker/journal/queue payload after each transition.

`make check` and `make check-race` passed on this revision.

Public `CLI_SAVEPOINT_SUPPORT` and SAVEPOINT statements remain unregistered until PR5.


apstndb:
Reviewed HEAD `1027cb60ba6ef9fc85c1c5ee1dbfc141f76a7708` against prerequisite `9e9ecd2756e1c445e28fb07fd9ebb64d6fa6fe0b`. Reconstruction needs the following corrections, in addition to the prerequisite #954 review:

1. **Compare the newly observed DML affected count.** [replaySQL](https://github.com/apstndb/spanner-mycli/blob/1027cb60ba6ef9fc85c1c5ee1dbfc141f76a7708/internal/mycli/savepoint_replay.go#L263) discards the collector's rowCount and calls FinishDML with the original `e.affected`. A fake-gRPC test records DML with count 1, changes only the server count to 2, and ROLLBACK TO still succeeds. Feed the new count into the fingerprint; mismatch must end the logical transaction and prevent candidate commit. Cover both zero/nonzero counts and returned rows. **Runtime-reproduced correctness defect.**

2. **Trim markers by marker index as well as journal position.** [rollbackToSavepointLocked](https://github.com/apstndb/spanner-mycli/blob/1027cb60ba6ef9fc85c1c5ee1dbfc141f76a7708/internal/mycli/savepoint_replay.go#L185) discards lookup's index. SAVEPOINT a; SAVEPOINT b; SAVEPOINT c with no intervening SQL followed by ROLLBACK TO a leaves b/c valid, because truncateAfter removes only larger journal positions. Keep the target and earlier markers, remove all later markers, including equal-position ones, and release their retained references/accounting. Exercise pending, RO and RW. **Runtime-reproduced semantic defect.**

3. **Bind heartbeat to its physical attempt, not only the logical owner/replacing flag.** [The heartbeat guard](https://github.com/apstndb/spanner-mycli/blob/1027cb60ba6ef9fc85c1c5ee1dbfc141f76a7708/internal/mycli/transaction_manager.go#L1265) runs after replacement releases tm.mu, when replacing is false and owner is intentionally unchanged. Using installBeforeAcquireBarrier, block an old heartbeat, complete ROLLBACK TO, then release the old heartbeat: it sends SELECT 1 on the new hb-txn-2 attempt. Capture/check an attempt token or the originating heartbeat generation/context under the lock; verify the old delayed tick emits no RPC while the restarted heartbeat works. **Runtime-reproduced concurrency defect with barriers, no sleeps.**

4. **Restrict reconstruction recovery to explicit RW and preserve RO lifecycle.** [shouldEnterRecoveryLocked](https://github.com/apstndb/spanner-mycli/blob/1027cb60ba6ef9fc85c1c5ee1dbfc141f76a7708/internal/mycli/savepoint_replay.go#L58) has no mode check. Begin RO, create marker a, then inject PermissionDenied for a query: the RO handle is discarded and NeedsRecovery becomes true. ROLLBACK TO a then reports success while leaving mode RO with a nil handle. This is neither the preserved RO snapshot nor a valid recovered transaction. Keep RO marker-only behavior on a valid handle and never revive a failed RO transaction; preserve existing non-replay error/close semantics and actually close any handle being terminally discarded. **Runtime-reproduced lifecycle defect.**

5. **Reserve marker bytes before flushing automatic DML.** [CreateSavepoint](https://github.com/apstndb/spanner-mycli/blob/1027cb60ba6ef9fc85c1c5ee1dbfc141f76a7708/internal/mycli/savepoint_replay.go#L138) flushes before addSavepoint reserves the name. With queued automatic DML, retainedBytes near the limit, and a valid long name, one ExecuteBatchDml RPC is observed before the marker budget error. Reserve the marker first, release that reservation on failed flush, and create the marker only after successful flush; local admission failure must leave the queue and owner intact. Coordinate with #954's complete batch reservation fix. **Runtime-reproduced ordering defect.**

6. **Use a separate bounded cleanup context after command cancellation.** Both old-attempt cleanup and candidate.Rollback receive the same command ctx ([candidate cleanup](https://github.com/apstndb/spanner-mycli/blob/1027cb60ba6ef9fc85c1c5ee1dbfc141f76a7708/internal/mycli/savepoint_replay.go#L213)). When replay times out/cancels, that context is already unusable for best-effort server rollback. The accepted design requires one deadline across replay plus a distinct bounded cleanup context. Add a cancellation-during-replay wire test verifying attempted rollback/terminal owner cleanup and no commit on the failed candidate; retain the underlying error cause for errors.Is where practical. **Source-backed cleanup contract gap; no managed-runtime cleanup claim.**

Independent fake-gRPC overlays reproduce items 1–5. Keep public support disabled until these cases and the complete accepted end-to-end matrix pass. Continue fixes in the existing stack, without a duplicate worker or merge.

apstndb:
Additional recovery-path finding at `1027cb60ba6ef9fc85c1c5ee1dbfc141f76a7708`: **owner/attempt binding must cover failure handling as well as receipt completion.**

[executeSQLImplWithQueryRunner calls HandleOwnerFailure on whichever owner is current](https://github.com/apstndb/spanner-mycli/blob/1027cb60ba6ef9fc85c1c5ee1dbfc141f76a7708/internal/mycli/execute_sql.go#L371). A deterministic wrapper probe creates iterator A, fully rolls back A, starts owner B and creates B's marker, then lets A's old iterator be consumed. The SDK correctly returns FailedPrecondition because A ended, but the wrapper discards B's physical handle and sets B to recovery-required. B had no failed operation.

When carrying #954's new captureToken into this layer, require the captured owner/attempt identity for **all** post-operation transaction error handling, including HandleOwnerFailure and ordinary aborted cleanup, not only finishQueryCapture. Note that finishQueryCapture clears pending, so a post-completion owner/attempt check must not depend on pending still matching. An unadmitted or intentionally independent single-use operation must not poison the active explicit owner either.

Add the above stale-error test through the real executeSQLImplWithQueryRunner wrapper, with an actual stopped/retired SDK iterator and a live replacement owner/marker. Assert the exact old-iterator error, B's unchanged handle/journal/recovery state, and no cleanup RPC targeting B. **Runtime-reproduced lifecycle defect.** This is part of the existing owner/attempt contract; continue in the same additive reconstruction fix, preserving dirty public-layer work.

apstndb:
One additional entrypoint must use the owner-bound error handling described above: **EXPLAIN ANALYZE SELECT, including CLI_QUERY_MODE=PROFILE.**

At `1027cb60ba6ef9fc85c1c5ee1dbfc141f76a7708`, [executeExplainAnalyze](https://github.com/apstndb/spanner-mycli/blob/1027cb60ba6ef9fc85c1c5ee1dbfc141f76a7708/internal/mycli/statements_explain_describe.go#L479) still returns through rollbackReadWriteIfAborted rather than the new recovery transition. Independent fake-gRPC probe: BEGIN RW; SAVEPOINT keep; inject PermissionDenied while executing EXPLAIN ANALYZE SELECT 1. The actual query fails with PermissionDenied, but NeedsRecovery remains false and the live RW attempt remains available. ABORTED follows the old full-retirement path instead of preserving the earlier marker.

Wire this query entrypoint through the same owner/attempt-bound completion and failure contract as ordinary SELECT. Test an actual failed stream after a completed marker, and verify both recovery rejection and successful ROLLBACK TO of the earlier prefix. **Runtime-reproduced missing recovery for PermissionDenied; ABORTED behavior is source-backed.** Please include this in the current reconstruction correction before exposing the public setting.

apstndb:
Re-reviewed `b68cfee27ed4d56ddd38bddc99aa2b4c8a882a3b` (including prerequisite #954 head 2e84f30). The original independent probes for changed DML count, equal-position markers, delayed heartbeat acquisition, RO handle preservation and marker admission before flush now pass. Cleanup now uses a separate bounded context and preserves the underlying error.

A new failure-path regression needs fixing: **the first SAVEPOINT panics if its automatic-DML flush fails.** [CreateSavepoint's error branch](https://github.com/apstndb/spanner-mycli/blob/b68cfee27ed4d56ddd38bddc99aa2b4c8a882a3b/internal/mycli/savepoint_replay.go#L161) dereferences tm.tc after flushAutomaticDMLLocked can retire it. Reproducer: BEGIN RW; enqueue one automatic DML; inject PermissionDenied for ExecuteBatchDml; CREATE first marker. There is no prior completed marker, so terminal rollback clears tm.tc and `tm.tc.replay.release(n)` panics. Keep the reservation owner/replay-state reference locally and release it safely regardless of logical retirement. Test both first-marker failure (terminal owner, no panic) and failure with an earlier marker (recovery owner, no new marker, correct bytes). **Runtime-reproduced crash.**

Two already-published follow-ups also still reproduce on this HEAD and remain required: [old iterator failure poisoning a replacement owner](https://github.com/apstndb/spanner-mycli/pull/955#issuecomment-5652516872) and [EXPLAIN ANALYZE missing recovery](https://github.com/apstndb/spanner-mycli/pull/955#issuecomment-5652529979). Apply the same owner/attempt check atomically with transaction error handling, rather than a check followed by separately locked cleanup.

Please also finish two narrow aspects of the existing heartbeat/cancellation requests:
- `startedAttempt` is currently read inside the new heartbeat goroutine. If that goroutine starts only after replacement, it can snapshot the replacement attempt. Check the originating heartbeat context's cancellation under the RPC-acquisition lock as well, or capture its generation before launching the goroutine. The existing delayed-acquire test does not cover delayed startup. **Source-backed scheduling gap.**
- `TestSavepointReplayCancellationCleansUpWithoutCommit` injects a server Canceled status while the command context is live, and accepts any rollback ID (including the old attempt). It therefore does not demonstrate the requested cleanup under an actually cancelled command context or candidate cleanup. Add a barrier that cancels after the candidate has an ID, then assert cleanup targets that candidate and no candidate commit. **Source-backed validation gap; the bounded-context production change itself is present.**

No acceptance yet; continue these corrections in the existing stack and carry #954's remaining fixes forward before the public layer.

apstndb:
Re-reviewed `cdbadb0a7dde3d1670b83b4506d625126cf1c740`. The original stale-iterator and EXPLAIN ANALYZE PermissionDenied reproductions now pass. `HandleOwnerFailure` correctly checks owner/attempt under its mutation lock and rejects nil tokens for independent operations.

One requested atomicity condition is still missing in the fallback: [`rollbackReadWriteIfAborted`](https://github.com/apstndb/spanner-mycli/blob/cdbadb0a7dde3d1670b83b4506d625126cf1c740/internal/mycli/execute_sql.go#L297) calls `tokenBelongsToCurrentOwner`, releases that read lock, calls `InReadWriteTransaction` under a separate lock, and then calls `RollbackReadWriteTransaction` under another lock. A completion with an Aborted error can pass the owner check for A and then roll back replacement B if a full rollback/begin occurs between the check and action. This is the check-then-act window specifically identified in the earlier review; passing the sequential stale-token test does not close it. **[P2] Source-backed concurrency defect, not claimed as a runtime reproduction.**

Move owner/attempt validation, mode/Aborted checks, and `RollbackReadWriteTransactionLocked` into one transaction-manager critical section, preserving the original error and the intentional single-use/disabled-mode policy. Avoid a new unlocked boolean helper as the safety boundary. The first-marker flush panic, heartbeat delayed-start guard, and actual-cancelled-context cleanup test remain tracked in the previous review while their fixes are queued; those bodies are not repeated here.

apstndb:
Execution receipt only — not a review verdict, not acceptance, and not a request for review replies.

Existing worker https://cursor.com/agents/bc-e4aa8f59-1f19-52a9-8598-f707b0c118e3 is in **local implementation** on reconstruction (not CI-wait, not a design blocker). Public run metadata still points at the old #952 / `cursor/savepoint-01-journal` label; that is a display gap, not a second lane.

**Current action:** applying remaining #955 reconstruction fixes on `/home/ubuntu/cohesion-923/.worktrees/savepoint-pr4` (`cursor/savepoint-04-reconstruction`). Published HEAD is still `fa139e332e54f2404d87382ffae66847ced0e19b`. This turn has not pushed. Last completed checkpoint left this worktree clean after `make check` / `make check-race`. Public layer `/home/ubuntu/cohesion-923/.worktrees/savepoint-pr5` (`cursor/savepoint-05-public` `7b97c54dca25c172401ad1b84538c3fa95b18d58`) is preserved and not mutated in this turn. #954 remains `98c331b1a6865ca770586b16247ed9c4a3cf3f4b`.

**Applying now** (consumed; no duplicate bodies): https://github.com/apstndb/spanner-mycli/pull/955#issuecomment-5652557578 — first-marker automatic-DML flush panic, heartbeat delayed-startup, cancelled-command cleanup test. Owner/attempt failure and EXPLAIN ANALYZE recovery are already on `cdbadb0` / `fa139e3`. The worker also folded atomic aborted-rollback cleanup into this #955 batch.

**Queued after this #955 batch (not in the current turn):** https://github.com/apstndb/spanner-mycli/pull/955#issuecomment-5652712374 (include in the same #955 push if not already in the current edits); then #956 https://github.com/apstndb/spanner-mycli/pull/956#issuecomment-5652631499 and https://github.com/apstndb/spanner-mycli/pull/956#issuecomment-5652694038 together; then the accepted e2e matrix. Public SAVEPOINT stays disabled.

No blocking error or design question. Cursor owns routine CI; Codex owns review resolution, semantic acceptance, and merge.

apstndb:
Re-reviewed `e0101db4f98b783d3c65ae01aa122b8bf794daa1`: the earlier independent reconstruction probes all pass and independent `make check` passed. The first-marker reservation release is now owner-safe; Aborted fallback validation and rollback share one lock; heartbeat generation is captured before launch and cancellation is checked at RPC acquisition. The committed cancellation test now targets a known candidate under an actually cancelled command context.

The remaining entrypoint audit found one additional omission: **[P2] PLAN-mode EXPLAIN/DESCRIBE SELECT failures bypass owner recovery.** [`RunAnalyzeQuery`](https://github.com/apstndb/spanner-mycli/blob/e0101db4f98b783d3c65ae01aa122b8bf794daa1/internal/mycli/transaction_manager.go#L1061) discards the operation token and returns `consumeRowIterDiscard` errors directly. `startOwnerSQLCaptureLocked` returns nil for PLAN mode, so simply passing its current token would still not associate the failure with the owner. The DML analyze path already goes through the locked RW failure handler; the SELECT path does not.

Runtime reproduction through `BuildStatement` / `Session.ExecuteStatement`: enabled explicit RW; successful SELECT 1; marker keep; inject Aborted (also reproduced with PermissionDenied); execute `EXPLAIN SELECT 2`. The exact RPC error is returned, but NeedsRecovery remains false and `CreateSavepoint("after_error")` succeeds. A subsequent COMMIT on the aborted physical attempt can terminate the owner instead of first requiring recovery. The successful-PLAN exclusion from replay does not exempt an owner RPC failure from the accepted recovery policy.

Keep successful PLAN operations out of the replay journal, while carrying owner/attempt identity (and the in-flight lifetime where applicable) through their collection/error path. Apply the same atomic failure boundary, preserve single-use/RO isolation, and test failed EXPLAIN/DESCRIBE SELECT with a prior marker plus successful ROLLBACK TO. Do not hash plans or manufacture a data replay entry for PLAN. Please fold this into the existing #955 correction lane and carry it forward; no separate PR is needed.

apstndb:
Independent re-review of `14693e6d02f8be760d088950ed6148403c3b23e3`: **no further actionable findings in this private reconstruction/recovery layer.**

The reported DML-count, equal-position marker, stale heartbeat/iterator, RO recovery, marker admission, bounded cleanup and first-marker failure issues are addressed. `e0101db` makes Aborted owner validation/cleanup atomic and captures heartbeat attempt identity before launch; `14693e6` carries PLAN-only owner identity through collection and recovery without journaling successful plans. The real cancellation test now observes and cleans up the actual candidate transaction.

All ten independent reconstruction probes pass, including public EXPLAIN with both Aborted and PermissionDenied. Independent `make check` and `make check-race` passed on this exact HEAD. Current-head CI checks passed; unresolved review-thread inventory is empty.

This accepts the scope of #955 only. Public command admission, buffered output completion and the final end-to-end acceptance matrix remain tracked on #956. No merge performed.

PR #956, opened 2026-09-13
Expose SAVEPOINT statements behind CLI_SAVEPOINT_SUPPORT

### LINKED ISSUE #404: feat: Support SAVEPOINT, RELEASE SAVEPOINT, and ROLLBACK TO SAVEPOINT

apstndb:
Fixes #404. Depends on #955 at `14693e6d02f8be760d088950ed6148403c3b23e3`.

Adds opt-in client-emulated `SAVEPOINT`, `ROLLBACK TO [SAVEPOINT]` and `RELEASE [SAVEPOINT]` commands for explicit transactions. `CLI_SAVEPOINT_SUPPORT` remains `DISABLED` by default and can change only while idle with no manual batch open.

Rollback replaces the physical read-write attempt, replays the recorded prefix, and validates typed ordered results, DML counts and Batch DML count vectors. Recovery preserves an earlier valid checkpoint after execution or output failures; when none remains, the transaction ends. Command identity prevents delayed output failures from affecting replacement attempts. Ordinary SET and transaction-scoped SET LOCAL retain their documented semantics.

Validation: `make check`, `make check-race`, exact-head CI, fake RPC tests for identity/replay/failure boundaries, and emulator tests for mixed DML, THEN RETURN, MUTATE/key ranges, manual/automatic batches, constraint recovery and committed contents. CLI tests cover buffered/streamed TABLE, CSV and JSONL output isolation, overlapping output completion, recovery admission and session-switch callback publication.

This is experimental replay, with a 16 MiB retained-payload budget; it does not preserve native locks/snapshots or prove equality of hidden reads and unreturned volatile writes. Large-stream enabled/disabled performance and managed-Spanner contention were not measured. See [SAVEPOINT semantics and limits](https://github.com/apstndb/spanner-mycli/blob/cursor/savepoint-05-public/docs/savepoint.md).


apstndb:
Independent review of `7b97c54dca25c172401ad1b84538c3fa95b18d58`: two additional public-surface findings, reproduced with the real `BuildStatement` -> `Session.ExecuteStatement` path and the fake Spanner RPC service.

1. **[P1] Enforce the recovery-required command boundary before state changes.** After ENABLED / BEGIN RW / SAVEPOINT keep and a PermissionDenied query, the transaction correctly reports NeedsRecovery. However, `SET CLI_VERBOSE = TRUE`, `SET LOCAL CLI_VERBOSE = TRUE`, `START BATCH DML`, `START BATCH DDL`, and `ABORT BATCH` all succeed. Starting a batch also makes ROLLBACK TO fail its manual-batch guard. [`Session.executeStatement`](https://github.com/apstndb/spanner-mycli/blob/7b97c54dca25c172401ad1b84538c3fa95b18d58/internal/mycli/session.go#L850) has no recovery admission policy; the scattered transaction-manager guards cannot cover these commands. Apply the accepted recovery allowlist before command side effects (ROLLBACK TO, full ROLLBACK/CLOSE, help and local inspection), with tests that rejected commands preserve variables, batch state and the logical owner. Inspect parameter/state-changing and SQL routes too; do not merely special-case these five strings.

2. **[P2] Reject support-mode changes while a manual batch is open.** `START BATCH DDL` or `START BATCH DML` while idle, followed by `SET CLI_SAVEPOINT_SUPPORT = 'ENABLED'`, succeeds. The accepted enablement contract requires idle **and no manual batch**. [`txnGuard`](https://github.com/apstndb/spanner-mycli/blob/7b97c54dca25c172401ad1b84538c3fa95b18d58/internal/mycli/var_registry.go#L151) checks only the transaction callback, so this setting needs the batch restriction at the appropriate public setter boundary. Test both batch modes and both direction changes; rejected SET must preserve the setting and batch.

Name parsing probes for parameters, paths, string literals, trailing input and 128/129 Unicode code points passed. The #954/#955 findings remain tracked on those PRs; carry their additive fixes into this head. Acceptance of the public feature remains pending those corrections and the complete end-to-end matrix.

apstndb:
Additional public-boundary finding at `7b97c54dca25c172401ad1b84538c3fa95b18d58`: **[P2] buffered CLI output failures bypass recovery and leave the failed command checkpointable.**

Reproduced through `Cli.executeStatement`: ENABLED; BEGIN RW; SAVEPOINT keep; execute SELECT 1 with TABLE, `CLI_TABLE_STREAMING=FALSE`, markdown codeblock enabled, and a writer returning a sentinel error. The CLI returns that exact error from `displayResult` / resultSink, but NeedsRecovery remains false, the physical attempt is usable, and the query is already in the replay journal. CSV's streaming writer path correctly enters recovery for the same writer failure, so the behavior depends on output mode. The documented receipt contract says formatter/pager failures are not checkpointable successes, and the accepted recovery matrix includes incomplete/cancelled output.

The collection/capture boundary finishes before buffered rendering: `executeWithBuffering` retains typed rows for later `printTableData`, and [`Cli.executeStatement`](https://github.com/apstndb/spanner-mycli/blob/7b97c54dca25c172401ad1b84538c3fa95b18d58/internal/mycli/cli.go#L640) currently returns display errors without transaction failure handling. Extend the enabled explicit-transaction command boundary so an actual buffered rendering/output failure after a marker cannot admit another SAVEPOINT or COMMIT without recovery. Preserve owner/attempt identity and do not retrospectively reconstruct a completed COMMIT or affect a replacement/independent transaction. The failed command must not be retained as a successful replay operation. Cover both successful end-to-end buffered output and an actual failing result sink; a row-iterator Write stub or CSV-only test does not cover this path.

This is separate from the recovery allowlist finding, which only restricts commands after recovery is entered. The exact failing probe used a TABLE result with a failing opening markdown fence to ensure the real sink reports the error. No external Spanner runtime or pager process was required.

apstndb:
One additional CLI-boundary finding at `7b97c54dca25c172401ad1b84538c3fa95b18d58`: **[P2] skip legacy client recreation for a recoverable SAVEPOINT abort.**

Reproduced through `Cli.executeStatement` with actual in-process Spanner clients: ENABLED; BEGIN RW; SAVEPOINT keep; SELECT returns Aborted. The owner correctly enters recovery, but [`Cli.executeStatement`](https://github.com/apstndb/spanner-mycli/blob/7b97c54dca25c172401ad1b84538c3fa95b18d58/internal/mycli/cli.go#L615) unconditionally calls RecreateClient for Aborted. A new client is constructed, `SetClient` rejects the still-live logical transaction, and the user receives an extra `cannot replace client during read-write transaction` error. The old owner/client are preserved and subsequent ROLLBACK TO succeeds; this is an unnecessary connection attempt and misleading secondary error, not demonstrated data loss.

Preserve the original Aborted error and the recoverable client/owner without running legacy recreation for this case. Keep existing terminal-abort recreation when capture is disabled, there is no checkpoint, or reconstruction/commit has ended the owner. Include the actual CLI path in the regression test; Session-only tests cannot exercise this branch. Please fold this into the current public-layer correction batch.

apstndb:
Allocation receipt only — not a review verdict, not acceptance, and not a request for replies.

Current #404 SAVEPOINT work stays on the existing worker https://cursor.com/agents/bc-e4aa8f59-1f19-52a9-8598-f707b0c118e3. Implementation, regression tests, and `docs/savepoint.md` are **not** split across workers.

Reason: remaining public-layer items share `internal/mycli/cli.go` and `session.go` (recovery allowlist, buffered-output recovery, RecreateClient skip), plus the same public tests and `docs/savepoint.md`. That worker already has dirty unpublished edits on `savepoint-pr5`. A second worktree would overlap those files.

Idle Cursor workers from other closed lanes are not reassigned. #954/`98c331b` and #955/`14693e6` stay accepted open bases. No extra publication authority; #956 stays draft until this batch plus exact-head CI, then Codex review.

Next checkpoint: one #956 push covering the current public-layer batch, then `make check` / `make check-race` and exact-head CI.

apstndb:
Re-reviewed `9d7bed0ed4738e279d42cba2fca09061acbdedcf`. Independent public recovery-admission/manual-batch tests and the actual-CLI Aborted/client test now pass. The buffered-output correction still has three reproducible issues:

1. **[P2] Carry the failing command's identity through display instead of reading the current owner's last capture.** [`handleBufferedOutputFailure`](https://github.com/apstndb/spanner-mycli/blob/9d7bed0ed4738e279d42cba2fca09061acbdedcf/internal/mycli/savepoint_replay.go#L89) reads `tm.tc.lastCommitted` at failure time, so `belongsToLocked` validates the newly selected token rather than the failing command. Reproduced with a writer barrier: SELECT 1 completes collection; before its writer returns an error, full ROLLBACK / BEGIN RW / SAVEPOINT next / SELECT 2 runs; the old SELECT 1 display error retracts SELECT 2 and puts the new owner into recovery. The same problem occurs after ROLLBACK TO replaces only the physical attempt. Pass a command-specific owner/attempt receipt to the display completion boundary (for example with the result/operation context); do not infer it from a mutable latest-operation slot. Rejected, independent and stale commands must not consume another command's completion state.

2. **[P2] Include successful Batch DML and MUTATE in the output completion boundary.** With ENABLED / BEGIN RW / SAVEPOINT keep, both `MUTATE T INSERT STRUCT(1 AS id)` and a successful manual `RUN BATCH` return the exact injected writer error when verbose TABLE output opens its markdown fence. However, NeedsRecovery remains false and each failed command stays in the journal. Only SQL/DML capture sets `lastCommitted`; batch/mutation completion does not. Apply the original failed-output contract across all supported owner operations, preserving independent operations and completed COMMIT semantics. Add actual `Cli.executeStatement` regression cases; a SELECT-only test does not cover these routes.

3. **[P2] Clear the removed journal slot before truncation.** [`retractLastSQL`](https://github.com/apstndb/spanner-mycli/blob/9d7bed0ed4738e279d42cba2fca09061acbdedcf/internal/mycli/savepoint_journal.go#L146) slices away the last entry and releases its budget without clearing the backing slot. After the existing buffered SELECT failure, the journal length is zero but its backing storage still contains the frozen SQL/parameter references. This repeats the retention issue already corrected in the other journal truncation paths. Release the actual references along with the accounting and test that boundary.

The original acceptance matrix is also still incomplete in committed tests: wire-level changed row order, equal-total/different Batch DML vectors, frozen replay request options/parameters/tags with next-owner slot behavior, no replay bytes/cache/metrics across supported output routes, and representative mixed-write/constraint-recovery emulator cases. The added nested-marker emulator test does not replace those cases. Root emulator probes already demonstrate the positive semantics, but retained regression tests were part of the accepted feature scope. Please provide test-name evidence for cases already covered and explicitly identify any proposed deviation instead of declaring the whole matrix complete.

Please keep these corrections in #956. Private layers #952-#955 remain accepted at their recorded heads; no merge is requested.

apstndb:
One further source-path check at `9d7bed0ed4738e279d42cba2fca09061acbdedcf` reproduced **[P2] a failed USE candidate replacing the live manual-batch guard callback**.

[`newConstructedSession`](https://github.com/apstndb/spanner-mycli/blob/9d7bed0ed4738e279d42cba2fca09061acbdedcf/internal/mycli/session.go#L422) immediately assigns `sysVars.inManualBatch = s.batch.IsActive`. USE/DETACH candidates share the live systemVariables and deliberately keep the existing transaction callbacks unbound until `adoptSession`. The new callback bypasses that boundary: construct a candidate, reject it during `switchSession` validation, retain the original Session, then START BATCH DML on that original Session. `SET CLI_SAVEPOINT_SUPPORT = 'ENABLED'` now succeeds because the callback reads the rejected candidate's empty batch. The independent probe uses the real constructor/switch failure path and public START BATCH/SET parsing; it fails with SET returning nil.

Bind the manual-batch callback at the same successful initial/adoption boundary as the other live session callbacks. Failed construction/validation must leave the previous callback intact. Add a failed-candidate regression plus successful initial/adoption checks, rather than fixing only the test Session factory. Keep this correction in the current public-layer batch.

apstndb:
Re-reviewed `c449acbd1aa2f76357ec837176179e6bb4fdf8f0`. The previously reported replacement-owner/attempt, MUTATE/manual Batch DML output, backing-slot retention, and failed-candidate callback probes now pass. The actual CLI output-isolation matrix also passes for buffered TABLE, streaming TABLE, CSV and JSONL.

One remaining **[P2] incomplete buffered output can still become a successful checkpoint within the same owner/attempt**. The new command token fixes replacement identity, but [`finishQueryCaptureLocked`](https://github.com/apstndb/spanner-mycli/blob/c449acbd1aa2f76357ec837176179e6bb4fdf8f0/internal/mycli/savepoint_capture.go#L204) releases inFlight and journals before display finishes. Reproduced through the actual CLI writer boundary: ENABLED; BEGIN RW; SAVEPOINT keep; SELECT 1 reaches buffered display; before its writer returns a sentinel failure, SELECT 2 and SAVEPOINT incomplete execute on the same attempt. The failure then enters recovery, but retractLast only examines SELECT 2, so SELECT 1 remains. ROLLBACK TO incomplete succeeds and replays that failed-output command. This violates the existing incomplete-output acceptance contract; it is a command-boundary concurrency test, not a claim that the interactive prompt itself runs commands concurrently.

Keep a command uncheckpointable through output completion, or invalidate its exact journal position plus any dependent suffix/markers when output fails. Do not identify entries by equal SQL text/payload size or batch length: distinct commands can share those values. A bounded writer-barrier test should prove that the intermediate checkpoint is rejected or invalidated and cannot later replay the failed command. Preserve the now-passing replacement-owner/attempt behavior and avoid hanging the writer under a reentrant transaction lock.

For the original retained regression matrix, the remaining specific gaps are small: `TestSavepointPublicStreamingOutputIsolation` allocates output buffers but never asserts replay emits zero bytes; add actual command/default-writer assertions with real-query positive controls across TABLE buffered/streaming, CSV and JSONL. `TestSavepointEmulatorMixedWritesAndConstraintRecovery` currently contains ordinary INSERT/UPDATE only; retain one representative mixed prefix with MUTATE INSERT, MUTATE DELETE KEY_RANGE, manual Batch DML and THEN RETURN, then verify actual committed contents after rollback/replay. `TestSavepointReplayUsesFrozenParamsOptionsAndTransactionTag` sets the next-owner tag only after replay/commit, which cannot detect replay consuming an already-set next-owner slot; set that slot before rollback and verify it survives into the next BEGIN, with later SQL using the current parameter/option values. Reuse current helpers/tests and avoid unrelated refactoring or an extra PR.

The four private PR verdicts are unchanged. This is targeted follow-up on the current public-layer correction, not acceptance or a merge request.

apstndb:
Allocation/execution receipt only — not a review verdict, not acceptance, and not a request for replies.

No second worker was assigned. Current #404 SAVEPOINT implementation stays on https://cursor.com/agents/bc-e4aa8f59-1f19-52a9-8598-f707b0c118e3. Idle workers from other closed lanes were not reassigned. The newest public review comments ([5652983392](https://github.com/apstndb/spanner-mycli/pull/956#issuecomment-5652983392) and [5653023066](https://github.com/apstndb/spanner-mycli/pull/956#issuecomment-5653023066)) are being handled together on this PR.

Published HEAD is `c449acbd1aa2f76357ec837176179e6bb4fdf8f0`. Exact-head CI on that commit is green (`test`, `race`, `lint`, `readme-sync`, `govulncheck`, cross-compile). The same worker is still active on one remaining CLI-path case from 5652983392: stale buffered display after `ROLLBACK TO` replaces only the physical attempt. #956 stays draft. Public SAVEPOINT stays disabled. No extra publication authority.

Already on `c449acb` (test names, not acceptance):
- command-specific display vs a replacement owner: `TestSavepointPublicBufferedOutputFailureIgnoresStaleCommand`
- Batch DML / MUTATE output completion: `TestSavepointPublicBufferedOutputFailureRetractsMutateAndBatch`
- retracted journal slot clear: `TestReplayStateRetractLastClearsBackingSlot`
- failed USE candidate batch callback: `TestSavepointPublicFailedUseCandidatePreservesManualBatchCallback`
- recovery allowlist: `TestSavepointPublicRecoveryAllowlistPreservesState`
- SET while a manual batch is open: `TestSavepointPublicSetRejectedInManualBatch`
- buffered TABLE markdown-fence failure: `TestSavepointPublicBufferedOutputFailureEntersRecovery`
- recoverable abort skips RecreateClient: `TestCli_executeStatement_abortedSkipsRecreateWhenSavepointRecoverable`
- row-order mismatch: `TestSavepointReplayQueryRowOrderMismatchEndsTransaction`
- equal-total Batch DML count-vector mismatch: `TestSavepointReplayBatchCountVectorMismatchEndsTransaction`
- frozen params/options/tags plus next-transaction tag: `TestSavepointReplayUsesFrozenParamsOptionsAndTransactionTag`
- streaming output/cache/byte isolation: `TestSavepointPublicStreamingOutputIsolation`
- emulator mixed-write + constraint recovery: `TestSavepointEmulatorMixedWritesAndConstraintRecovery`
- emulator nested/pending markers: `TestSavepointEmulatorNestedReleaseAndPending`
- emulator insert/savepoint/rollback commit contents: `TestSavepointEmulatorCommitContents`

Still pending: the `ROLLBACK TO` physical-attempt stale-display CLI case above, then Codex review of the resulting head. No merge requested.

Next checkpoint: remaining CLI test on #956 if unpublished, exact-head CI on that SHA, then Codex review.

apstndb:
At `4a06d4415fe1b5e84aba1cac195c190de88ae15a`, the same-attempt suffix/marker probe now passes; the retained mixed-write, frozen-slot and multi-format output tests are materially improved.

One remaining **[P2] end the physical attempt when a buffered write failure leaves no valid recovery marker**. Reproduced independently through the actual CLI, without concurrency: ENABLED; BEGIN RW; execute INSERT, MUTATE, or manual RUN BATCH before the first SAVEPOINT, with verbose buffered TABLE/markdown output and a sentinel-failing writer. Each returns the sentinel error, retracts its journal entry, but leaves the physical RW transaction alive. A following SAVEPOINT succeeds with the failed write omitted from its journal.

[`handleBufferedOutputFailure`](https://github.com/apstndb/spanner-mycli/blob/4a06d4415fe1b5e84aba1cac195c190de88ae15a/internal/mycli/savepoint_replay.go#L92) only enters recovery when a marker remains, otherwise calling rollbackReadWriteIfAbortedLocked. An ordinary writer error is not Aborted, so this path leaves physical writes and replay history inconsistent. The same empty-marker condition can occur when suffix invalidation removes the only marker. The existing accepted contract says errors without a checkpoint preserve terminal rollback behavior and a new savepoint must not include an unrecorded failed operation.

For a current captured write whose output fails, preserve an earlier valid checkpoint when one exists; otherwise roll back and retire the logical/physical attempt. Reuse the existing owner failure/retirement path as appropriate, including SET LOCAL restoration, and preserve stale/replacement and already-committed implicit transaction behavior. Add no-prior-marker cases for INSERT, MUTATE and RUN BATCH, plus invalidation of the only later marker. The three independent no-marker probes fail on this head; all earlier independent public probes pass. Keep this small correction in #956; no new scope or PR is needed.

apstndb:
Re-reviewed `558e6f7039d77e1a8b682ab55fea166d21e8003c`: no-marker INSERT/MUTATE/RUN BATCH now retire correctly, including immediate SET LOCAL restoration. All earlier independent public probes and root `make check` pass.

**[P2] Preserve a surviving earlier checkpoint when another same-attempt display completion fails after recovery has already begun.** The new call to handleOwnerFailureLocked uses shouldEnterRecoveryLocked, which is false when recoveryRequired is already set; its fallback therefore retires the owner even if valid earlier markers remain.

Reproduced with the actual CLI writer boundary: ENABLED; BEGIN RW; SAVEPOINT keep; SELECT 1 reaches buffered display; before it finishes, SELECT 2 executes and its writer fails, entering recovery. Then SELECT 1's writer fails. Its exact entry is correctly retracted, and keep still denotes the valid empty prefix, but the second completion fully retires the owner. ROLLBACK TO keep is no longer possible. This extends the same-attempt overlapping-output scenario already covered in the preceding correction; it is not a claim about concurrent interactive prompt execution.

Distinguish “a valid recovery marker remains” from “first transition into recovery.” When valid markers remain after exact-ID invalidation, keep the recovery owner even if its physical attempt was already discarded; terminate only when no valid marker remains. Also check the opposite completion order, where an earlier failed output has already removed a later command's journal ID: that later completion must not consume another command's state. Cover both orders with a bounded barrier test in the same correction batch. Preserve the passing no-marker/only-invalidated-marker termination cases and stale replacement-owner/attempt cases. No broader refactor, new lane or extra PR is needed.

apstndb:
Reviewed `7011a941dcb7295ea56df4afcc6c34b217fcf647`: **no further actionable findings in the accepted SAVEPOINT scope**.

All previously reported public-layer findings are addressed. Independent probes pass for actual CLI recovery/admission and client-abort handling, output failures with no/earlier/invalidated markers, both stale replacement cases, same-attempt overlapping output, immediate SET LOCAL restoration, and zero replay bytes/cache changes across buffered TABLE, streamed TABLE, CSV and JSONL. Independent emulator checks verify actual committed contents for mixed DML/THEN RETURN/MUTATE/key ranges/manual batch, constraint recovery, automatic queues and frozen parameters.

Root `make check` and `make check-race` both exit 0 on this exact head (macOS, Go 1.26.8, Colima-backed emulator). All substantive current-head GitHub checks pass. No unresolved review threads or pending review drafts remain. The earlier #952-#955 acceptance verdicts still match their unchanged heads.

Validation covers the fake RPC and emulator environments; managed-Spanner contention and a large-stream enabled/disabled performance comparison remain unmeasured. The documented default-disabled client-emulation boundary is retained. No merge is performed or requested by this verdict.

PR #979, opened 2026-09-14
Verify expected update counts when automatic DML batches flush

apstndb:
## Summary

Adds `AUTO_BATCH_DML_UPDATE_COUNT` (INT64, default 1) and `AUTO_BATCH_DML_UPDATE_COUNT_VERIFICATION` (BOOL, default FALSE) for existing automatic DML batching.

Verification is opt-in so current arbitrary UPDATE/DELETE statements keep succeeding. Each queue entry freezes its expected-count policy at enqueue. A successful BatchUpdate compares each enabled actual count before a journal receipt; `[1,1]` vs `[0,2]` fails at statement 1. RPC and partial BatchUpdate errors keep their original cause. Mismatch uses the existing SAVEPOINT failure/recovery path. Replay still checks journaled actual counts after verification is turned off.

Manual, THEN RETURN, unbuffered, and partitioned DML are unchanged. No TRANSACTION_TIMEOUT / #482 owner-deadline edits.

Fixes #401

## Test plan

- [x] Focused short tests for per-entry match/mismatch, zero rows, frozen policy, RPC failure, SAVEPOINT recovery, SELECT/SAVEPOINT/RUN BATCH/COMMIT flush, enqueue presentation, and replay actual-count check
- [x] `make check` before push
- [x] `make check-race` for flush/owner lifecycle
- [x] Coverage of `transaction_manager_auto_dml.go` at or above 80%

<div><a href="https://cursor.com/agents/bc-475ba374-19f8-51ae-9dbe-d859196a47ce?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-475ba374-19f8-51ae-9dbe-d859196a47ce&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

Issue #357, opened 2025-06-25
Expire user-idle transactions independently of keepalive traffic


apstndb:
## Scope

Implement a CLI-owned idle-transaction policy. This remains distinct from TRANSACTION_TIMEOUT (#482), which limits total transaction duration, and from the KEEP_TRANSACTION_ALIVE configuration plan. #403 was closed because the legacy SDK session-pool removal option is a no-op; do not revive it.

## Implementation contract

- Add CLI_IDLE_TRANSACTION_TIMEOUT and --idle-transaction-timeout with explicit duration strings (for example 60s or 5m), consistent with --timeout. Default NULL/0 is disabled; reject negative/overflow. Do not introduce a default 60-second expiry.

- Track completed admitted user/database work on the logical owner, including successful buffered DML/MUTATE and SAVEPOINT operations. Invalid syntax/rejected admission does not count. Heartbeat RPCs and client-only polling/SHOW/SET do not reset idle. A successful explicit BEGIN RW/RO acquiring a server transaction starts an initial quiet interval at completion; a resource-free pending BEGIN need not arm until work. A constructor or heartbeat later does not reset idle.

- Hold idle expiry through the user RPC, complete iterator consumption and associated result rendering/pager completion. Idle is not a deadline on RPC contexts; TRANSACTION_TIMEOUT retains total-budget cancellation. Capture policy at BEGIN and preserve it across SAVEPOINT replacement/recovery. Session SET affects the next owner; SET LOCAL may change capture before admitted user work and then freezes.

- Enforce idle expiry even when heartbeat is disabled. Reuse the lifecycle/cancellation infrastructure, but do not make the idle timer's existence depend on keepalive being enabled.

- On expiry roll back/retire only the same owner, restore SET LOCAL at the existing safe barrier and report a one-time subsequent-command error without executing that command. Explicit ROLLBACK/CLOSE/BEGIN/USE/DETACH acknowledge prior expiry as teardown/new-owner intent. A delayed callback must never close a later owner; timers must not invoke registry setters. Test final-frame races and total-timeout precedence.

## Acceptance

- [ ] Deterministic tests cover enabled/disabled keepalive, active long query, user-idle interval, recovery, rollback, session switch and stale timer callbacks.

- [ ] No fixed sleep substitutes for synchronization; all lifecycle goroutines exit.

Related: #342 (CLI options), #482 (total deadline), #403 (closed obsolete SDK mechanism).

## Verification and delivery

[...]

Issue #293, opened 2025-06-17
Implement bounded ABORTED retries using the SAVEPOINT replay machinery


apstndb:
## Updated baseline

RETRY_ABORTS_INTERNALLY remains a placeholder. Since v0.35.0-alpha.2, transactionContext already owns a bounded journal, frozen parameters, typed ordered result fingerprints, automatic DML and SET LOCAL undo. The earlier suggestion to build a new replay engine is obsolete.

Both reference drivers default internal retry to true. Keep that as the compatibility proposal, but resolve the enablement/overhead contract in a design PR before silently enabling capture for all transactions. Preserve the distinction between driver-level logical retry and SDK RPC/stream retries; disabling this property must not claim to disable every SDK retry.

## Bounded implementation phases

1. Design the property/default and retry budget, then ship one-statement implicit RW retry as a complete PR. Classify ABORTED precisely and preserve parameters/options; never retry ambiguous commit outcomes.

2. Add explicit logical-transaction retry through savepoint_journal/replay/receipt helpers, not a database/sql migration or second journal. Capture from BEGIN when retry is enabled even when no SAVEPOINT marker exists.

3. Integrate observation completion: streamed output already delivered must not be duplicated, failed/unobserved output is not a successful receipt, and changed typed rows/order/counts must terminate retry.

## Acceptance

- [ ] Bound attempts by deadline/cancellation and a documented retry policy; no endless busy loop.

- [ ] Preserve the same logical owner, original transaction deadline, SET LOCAL undo, frozen constructor options and correct heartbeat attempt identity.

- [ ] Cover ABORTED on first statement, after reads, on automatic/manual Batch DML and on Commit; cover mismatch, partial output, non-ABORTED failure, cancellation and journal limit.

- [ ] Document hidden reads/unreturned volatile writes and opt-in/default overhead. Decide SET LOCAL eligibility before activation; do not discard this question because the old accessor example used a simple bool.

- [ ] Integrate with #482; design it first or carry a shared remaining-deadline interface so retry never renews the transaction budget.

[...]

PR #980, opened 2026-09-14
feat: add TRANSACTION_TIMEOUT logical deadline

### LINKED ISSUE #482: Apply TRANSACTION_TIMEOUT to the logical transaction across replay
## Current scope

Implement a logical read/write transaction deadline distinct from STATEMENT_TIMEOUT and user-idle timeout (#357). The old SET LOCAL prerequisite #400 is complete. The existing owner/attempt separation from SAVEPOINT must retain one budget across replacement.

## Contract to implement

- TRANSACTION_TIMEOUT accepts a duration or NULL; zero/NULL means no additional transaction deadline. Snapshot the configured duration for a logical transaction.

- Define the clock explicitly: start at its first actual database operation, not a client-only BEGIN/SHOW, while preserving the originally selected duration. Do not reset it on a later statement, ABORTED retry or ROLLBACK TO.

- A pending SET LOCAL may select the duration before first database use; reject attempts to extend/change the budget after activation. Session SET after BEGIN applies to a later logical owner, not the current one.

- Every SQL/Batch DML/commit/replay RPC receives the minimum of caller, statement and remaining transaction deadline. A client-only SHOW must not activate or renew it.

- On expiry cancel/retire the matching logical owner, stop its heartbeat and restore SET LOCAL exactly once. No stale timer may cancel a replacement owner.

## Design clarification (2026-09-14)

- Verify against the current commit and pinned SDK, not a stale coordinator checkout. At the audited baseline, mycli uses Go Spanner v1.95.0 ReadWriteStmtBasedTransaction. Its default constructor normally sends an explicit BeginTransaction RPC; generic pending BEGIN is client-only. The statement-based wrapper overrides the embedded callback-based transaction's begin decision. Start the budget before the actual transaction RPC, including constructor BeginTransaction when applicable. Merely taking a multiplexed session or buffering DML without a transaction RPC does not start it. A buffered-DML route that activates the owner first can nevertheless already have issued BeginTransaction.
- A deadline must cancel an in-flight RPC without first waiting for TransactionManager.mu: some existing RPC paths hold that mutex. Retire only the matching logical owner under the mutex after cancellation. A captured owner pointer and unchanged total budget must survive physical attempt replacement.
- Timer goroutines must not mutate the live registry through Registry.Set. Detach the owner's LOCAL undo for restoration at a serialized session/CLI safe point. Drain expired-owner undo before a later statement reads execution defaults or creates a new owner, and on Close; the ordinary end-of-statement defer also handles an in-flight operation returning after cancellation. Document this safe-point restoration boundary. A naked inExec boolean, even atomic, is not an atomic check-and-restore protocol.
- Test cancellation of an RPC while the transaction mutex is held, idle expiry followed immediately by BEGIN for a new owner, explicit-constructor failure, delayed old-owner callbacks, and first activation through automatic DML. Prove errors/ownership, not merely elapsed time. Preserve the existing failed-construction and SAVEPOINT recovery invariants.

## Acceptance

- [ ] Tests cover pending idle time, cumulative statements, long/streamed reads, automatic DML, commit, replay, shorter caller context and cancellation.

- [ ] Expiry ends the correct owner and retained attempts cannot restart its budget; no fixed sleeps are used as synchronization.

- [ ] Document distinction from #357 and interaction with #293.

References: [Go transaction timeout](https://github.com/googleapis/go-sql-spanner/blob/28b0df2646750c7aab56da320fe6fc945b30ec86/connection_properties.go#L494), [Java property](https://github.com/googleapis/google-cloud-java/blob/dd2c41a0e2ec05d39b93f7d94169d596bb7a5483/java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionProperties.java#L687), mycli session_transaction_context.go and transaction_manager.go.

## Verification and delivery

Plan refreshed on 2026-09-14 against `a8fa2418b65af9b889cf92e02190366b3e6e02ea`. This body supersedes obsolete implementation suggestions in earlier comments; history is retained. Tracked with #47.

For each implementation PR, capture focused test output to a file and check the exit code, then run `make check` before pushing; run `make check-race` for transaction concurrency/lifecycle changes. Keep the coverage gate at 80%. Update user docs and the compatibility matrix with the resulting behavior. No database/sql migration or unrelated dependency upgrade.

apstndb:
## Summary

Implements `TRANSACTION_TIMEOUT` as a logical read/write deadline for #482.

- Duration or `NULL`; `NULL`/0 means no additional transaction deadline.
- The duration is captured for the logical owner. Pending `SET LOCAL` may select it before the first real database RPC; changing it after the budget starts is rejected. Session `SET` or `RESET` after `BEGIN` applies to a later owner.
- The single total budget starts at the first real database RPC, including constructor `BeginTransaction` on pinned Go Spanner v1.95.0 `ReadWriteStmtBasedTransaction`. Client-only `BEGIN`/`SHOW` and buffering automatic DML do not start it.
- The deadline is preserved across physical reconstruction. SQL, Batch DML, commit, and replay RPCs receive the remaining budget together with the caller and `STATEMENT_TIMEOUT` deadlines.
- Expiry cancels in-flight RPCs without waiting for the transaction mutex, retires only the matching owner, stops its heartbeat, and restores `SET LOCAL` at the serialized session safe point (start/end of `ExecuteStatement`, and `Close`). Timer goroutines never call `Registry.Set`.
- Failed constructor keeps pending identity, undo, and an already-armed budget. Idle constructor failure still leaves no owner.
- Distinct from unimplemented user-idle expiry (#357). ABORTED retries (#293) are not implemented and are not added here.

Does not change automatic-DML queue/flush signatures owned by #979/#401.

Fixes #482

## Test plan

- [x] Focused unit tests for capture, SET LOCAL vs session SET/RESET, constructor start, failed construction, mutex-held cancel, idle expiry then BEGIN, delayed old-owner callback, SAVEPOINT budget reuse, automatic DML, NULL/0, shorter caller context
- [x] `go test -short ./...`, `golangci-lint run`, `make fmt-check` (exit 0)
- [x] `make check-race` (exit 0)
- [x] `go test -short -coverpkg=./internal/mycli ./internal/mycli` coverage 82.6% (floor 80%)
- [ ] Hosted CI `make test` / emulator (this VM has no Docker; local `make check` failed only on integration container startup)

<div><a href="https://cursor.com/agents/bc-ac5e54c0-507b-5981-b4cc-ca2b5fe3a7a6?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-ac5e54c0-507b-5981-b4cc-ca2b5fe3a7a6&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


apstndb:
Reviewed HEAD `051c9032c31cea94aeaaee251e271e826cce0a78`. Two blocking findings are independently reproduced with the real client connected to the existing local fake gRPC server:

1. **[P1] Pass the remaining deadline to every Batch DML RPC.** `RunInNewOrExistRwTxLocked` shortens its own `ctx`, but its callback still captures the caller's original context, so `executeBatchDML` sends that original context to `BatchUpdateWithOptions`. The automatic flush path also sends the unbound context except at one caller. With a 20-second caller and a 2-second transaction, the actual server-side Batch DML context had the 20-second deadline in both manual-batch and `FlushAutomaticDML` probes. These RPCs hold the transaction mutex, so the expiry watcher cannot retire the owner to compensate. Bind the context at the actual RPC boundary (or explicitly pass the bound context into the callback) for every route, including RUN BATCH and flush-before-read. Add actual fake-RPC deadline assertions and an in-flight cancellation/owner-retirement regression; merely checking the stored owner deadline or a COMMIT flush does not cover these paths. Locations: [callback dispatch](https://github.com/apstndb/spanner-mycli/blob/051c9032c31cea94aeaaee251e271e826cce0a78/internal/mycli/transaction_manager.go#L1513), [manual Batch DML](https://github.com/apstndb/spanner-mycli/blob/051c9032c31cea94aeaaee251e271e826cce0a78/internal/mycli/execute_dml.go#L139), [automatic flush](https://github.com/apstndb/spanner-mycli/blob/051c9032c31cea94aeaaee251e271e826cce0a78/internal/mycli/transaction_manager_auto_dml.go#L143).

2. **[P2] Freeze the selected duration after first database use even when it is unlimited.** The freeze check uses a nonzero deadline, but NULL/0 never sets one. Starting a transaction with the default NULL, observing its constructor BeginTransaction RPC, and executing `SET LOCAL TRANSACTION_TIMEOUT = '1h'` succeeds instead of returning the frozen-owner error. A later operation can therefore start a new budget on an already active owner, contrary to the pending-only selection contract. Track first-use/activation independently of whether a timer exists, preserve it across failed construction/replacement as specified, and cover both NULL and zero. Location: [freeze and arming](https://github.com/apstndb/spanner-mycli/blob/051c9032c31cea94aeaaee251e271e826cce0a78/internal/mycli/transaction_timeout.go#L71).

The independent probes failed with those exact causes on both `586684b` and the current merged HEAD. Review remains open while these are fixed.

The required pre-push gate also remains unmet: the receipt says `make check` exited 2 because Docker was absent, followed by another push. Prepare a working Docker/emulator environment or reuse the already demonstrated full-validation route, then run the full `make check` successfully before every further push, plus `make check-race` for these lifecycle fixes. Capture commands, exits and exact source identity. Short tests/lint and later hosted CI are not a substitute for that repository requirement. If environment recovery truly blocks the lane after concrete attempts, report the evidence before publishing again. Preserve all published history and fix additively; no reset, rebase or force-push is authorized.

apstndb:
Additional blocking finding on the same reviewed HEAD `051c9032c31cea94aeaaee251e271e826cce0a78`:

**[P1] Make expiry retirement and the statement-entry restoration boundary atomic.** `executeStatement` checks/drains LOCAL restoration only once at entry. If the timeout callback retires owner A immediately after that check, the statement can create owner B before A's detached undo is drained. The end-of-statement check then sees B and skips restoration, so B inherits A's expired LOCAL settings.

I reproduced this deterministically with an overlay hook immediately after the new entry drain, using the existing expiry callback before the actual `BEGIN` handler. A had session timeout 1h, LOCAL timeout 2h and LOCAL `CLI_VERBOSE=TRUE`; after A's expiry and B's BEGIN, verbose was still TRUE and B captured 2h rather than the restored 1h. This is a forced callback interleaving, not a wall-clock timing claim. No production behavior was changed by the probe beyond scheduling that callback at the existing race window.

Please implement an atomic handoff under the lifecycle synchronization so an expired owner's undo is restored before another statement reads defaults or installs a new owner. Registry setters must remain outside timer goroutines, and an unsynchronized/standalone execution flag or another one-shot drain would leave the same race. Add this exact ordering regression alongside expiry-before-entry and in-flight cancellation cases. Location: [entry restoration](https://github.com/apstndb/spanner-mycli/blob/051c9032c31cea94aeaaee251e271e826cce0a78/internal/mycli/session.go#L860), [timer retirement](https://github.com/apstndb/spanner-mycli/blob/051c9032c31cea94aeaaee251e271e826cce0a78/internal/mycli/transaction_timeout.go#L133).

Keep the same additive PR and the previously requested full `make check` plus race gate before publishing the fixes.

apstndb:
6e3308e1f59a7789ae891a102b586ba5f856df6e

apstndb:
Review update for exact HEAD `6e3308e1f59a7789ae891a102b586ba5f856df6e`:

The independent actual-RPC Batch DML deadline probes and NULL/0 first-use rejection now pass. The original timeout/verbose owner-install regression also passes. Thank you for keeping the published history additive.

**[P1] The expiry/LOCAL restoration boundary is still incomplete.** I reproduced both remaining cases with the new `afterEntryRestore` hook calling `watchTransactionDeadline(ownerA, expiredContext())`, then the actual statement handler:

- Session `RPC_PRIORITY=LOW`, A has `SET LOCAL RPC_PRIORITY=HIGH`; expire A after entry restoration, then execute `BEGIN`. B captures HIGH although the registry is restored to LOW. `BeginPendingTransaction` resolves priority/isolation before `withOwnerInstallAfterRestore` restores the registry ([lines 702 onward](https://github.com/apstndb/spanner-mycli/blob/6e3308e1f59a7789ae891a102b586ba5f856df6e/internal/mycli/transaction_manager.go#L702)).
- Session timeout1h, A LOCAL timeout2h; expire A at the same boundary, then execute ordinary `SET TRANSACTION_TIMEOUT = '3h'`. SET reports success, but the post-statement drain overwrites it with1h. Guarding only owner constructors cannot protect successful session setters or statement defaults read before owner installation ([statement entry/defaults](https://github.com/apstndb/spanner-mycli/blob/6e3308e1f59a7789ae891a102b586ba5f856df6e/internal/mycli/session.go#L860)).

These are deterministic forced-interleaving reproductions on this HEAD, not elapsed-time claims. Please fix the shared statement/retirement protocol, cover ordinary SET plus frozen priority/isolation and statement-default reads, and retain constructor/in-flight cancellation and nested execution coverage. Do not hold a non-reentrant statement mutex across an RPC or move Registry.Set into timer callbacks. A synchronized, owner-aware deferred-retirement protocol is one possible design; choose the smallest protocol with a clear ordering argument. Moving two default reads alone leaves the successful SET regression.

Validation remains full `make check` before every further push and `make check-race`, with full source identity and whether Docker integration actually ran. Please return the relevant successful gate summary, since the last progress text questioned unit-only execution. Current hosted CI is green; this source/runtime finding still blocks acceptance. Preserve published history and use additive fixes only.

apstndb:
8cd56b7be9b2710059b38e2a79790b62eb4c7ca8

apstndb:
Review update for exact HEAD `8cd56b7be9b2710059b38e2a79790b62eb4c7ca8`:

All earlier independent probes now pass, including the actual Batch DML deadlines, NULL/0 freeze, expiry before BEGIN with LOCAL priority, and ordinary session SET preservation. The reported uncached full/Docker gate receipt also resolves the earlier unit-only uncertainty.

**[P2] Close the final statement-depth/expiry handoff atomically.** `leaveStatement` first calls `syncExpiredOwnerRestore`, then separately locks and decrements depth. If the timeout callback runs between those operations, it sees depth1, sets `expirePending`, stops its watcher and leaves undo attached. `leaveStatement` then changes depth to0 and returns without another retirement. The now-idle owner stays live/expire-pending and retains LOCAL values and queued work until a later command, although its only expiry watcher has already exited.

I reproduced this on this HEAD with one test-only overlay callback immediately after the final `syncExpiredOwnerRestore` in `leaveStatement`, before the depth decrement. It invokes the existing expiry callback for A while an actual `SHOW VARIABLE CLI_VERBOSE` finishes. After return: depth0, non-nil owner with expirePending=true, CLI_VERBOSE still TRUE rather than restored FALSE. This is a forced legal callback ordering; there is no production state mutation in the probe beyond invoking that callback at the existing gap.

Please make the final frame-depth transition and decision to retire an expired matching owner a single tm.mu critical section, then restore detached registry entries outside the lock. Preserve nested-frame semantics and the rule that timer callbacks do not touch the registry. The invariant at that boundary should rule out depth0 with an already-processed expiry still pending and no watcher left to retire it. Add the exact final-frame interleaving regression; another unsynchronized one-shot drain alone simply moves this gap. Location: [leaveStatement](https://github.com/apstndb/spanner-mycli/blob/8cd56b7be9b2710059b38e2a79790b62eb4c7ca8/internal/mycli/transaction_manager.go#L396).

This is a small completion of the new protocol; keep it additive and run full make check plus make check-race before pushing. No merge or history rewrite by the worker.

apstndb:
8d375df628ec20418cbfe96189aea642be71970a

apstndb:
Independent re-review of exact HEAD `8d375df628ec20418cbfe96189aea642be71970a`: **no remaining actionable findings in the reviewed scope**.

The earlier findings are resolved additively: `9f05be0`/`6e3308e` bind the actual manual/automatic/replay Batch DML RPCs and track first use independently of NULL/zero; `8cd56b7` preserves expired LOCAL restoration across BEGIN/default capture and ordinary SET; `8d375df` makes final-frame depth/retirement atomic while keeping registry restoration outside the transaction mutex and timer callbacks. Source review also covers owner identity, nested statement frames, reconstruction and cleanup ordering.

On this exact HEAD, my independent fake-gRPC/deadline/forced-interleaving probes and the timeout test suite pass, including every previously failing reproduction. Independent full `make check` passed locally with Docker-backed integration, lint and format checks. Current hosted race and all other applicable CI checks pass; required-check wait passed, the branch is mergeable, and no unresolved inline or review-level feedback remains. The implementation worker also supplied successful full/race gate receipts for this source.

This acceptance concerns the logical read/write timeout and its integration; it does not add user-idle expiry or automatic ABORTED retry, and local fake/emulator tests are not a claim of managed-service validation.

Issue #402, opened 2025-07-19
Define DDL behavior inside logical transactions


apstndb:
## Scope

Implement an explicit runtime CLI_DDL_IN_TRANSACTION_MODE policy with FAIL (default), ALLOW_IN_EMPTY_TRANSACTION and AUTO_COMMIT_TRANSACTION, following Java's DdlInTransactionMode where appropriate. Preserve the fact that Spanner DDL is non-transactional; no mode can roll it back through SAVEPOINT.

## Implementation

- Read the existing DDL and transaction admission paths before editing. Do not wire a nonexistent NewTransactionOptionsBuilder from old comments; the current code uses transactionOptions and transactionContext.

- Capture the policy for the logical transaction; define permitted SET/SET LOCAL timing before activation and reject changes that could unexpectedly commit an existing owner.

- Define empty precisely, including pending BEGIN, reads already performed, buffered mutations, automatic/manual DML and savepoint markers. Checking only whether an SDK handle exists is insufficient.

- AUTO_COMMIT_TRANSACTION must successfully commit all pending work before issuing DDL. Commit failure, recovery-required state or local validation failure must prevent the Admin RPC.

- Specify retirement, SET LOCAL restoration, schema-cache invalidation and what is reported when commit succeeds but DDL fails. Coordinate with AUTOCOMMIT #83 and async modes #485.

## Acceptance

- [ ] A design table covers pending, empty/nonempty RW, RO, manual batch, recovery, successful commit then failed DDL, and SAVEPOINT attempts.

- [ ] Fake RPC tests prove ordering and absence of DDL after failed commit; emulator tests assert committed data cannot be rolled back by later DDL/SAVEPOINT failure.

Reference: [Java DdlInTransactionMode](https://github.com/googleapis/google-cloud-java/blob/dd2c41a0e2ec05d39b93f7d94169d596bb7a5483/java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/DdlInTransactionMode.java), [mycli DDL execution](https://github.com/apstndb/spanner-mycli/blob/a8fa2418b65af9b889cf92e02190366b3e6e02ea/internal/mycli/execute_ddl.go#L63).

## Verification and delivery

Plan refreshed on 2026-09-14 against `a8fa2418b65af9b889cf92e02190366b3e6e02ea`. This body supersedes obsolete implementation suggestions in earlier comments; history is retained. Tracked with #47.

[...]

PR #986, opened 2026-09-14
feat: opt-in mutation-limit DML fallback (#964)

### LINKED ISSUE #964: Design opt-in transactional DML fallback after mutation-limit failure
## Audited baseline

Audited 2026-09-14 against spanner-mycli `a8fa2418b65af9b889cf92e02190366b3e6e02ea` (v0.35.0-alpha.2). Source-backed unless explicitly marked runtime-reproduced. Reference versions are pinned in the links. Part of #47. Implementation wave 3 (design first; not a ready implementation ticket).

Java exposes TRANSACTIONAL_WITH_FALLBACK_TO_PARTITIONED_NON_ATOMIC and [catches TransactionMutationLimitExceededException specifically](https://github.com/googleapis/google-cloud-java/blob/dd2c41a0e2ec05d39b93f7d94169d596bb7a5483/java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/SingleUseTransaction.java#L597). mycli's enum has only TRANSACTIONAL and PARTITIONED_NON_ATOMIC. The Go SQL driver has only those two modes too.

## Scope and approach

1. Specify the explicit third AUTOCOMMIT_DML_MODE value without changing the default. Explain that fallback weakens atomicity, can partially commit on later failure, and returns a lower-bound count.
2. Limit eligibility to one implicit/autocommit DML statement, with no returned rows, manual/automatic batch, explicit transaction or SAVEPOINT owner.
3. Establish a source/runtime-backed Go error classifier proving mutation-limit failure and an uncommitted original attempt. Never fall back for generic RESOURCE_EXHAUSTED, ABORTED, timeout or ambiguous Commit failure.
4. Freeze parameters and relevant request options. Reject unsupported/non-partitionable statements through the normal PDML path; retain both original and fallback error information.
5. Do not implement if the Go SDK cannot distinguish the required failure safely; retain an actionable upstream/evidence blocker instead of a message-only heuristic.

## Acceptance and validation

- [ ] Design explicitly lists positive and negative error cases and data-integrity implications.
- [ ] Fake RPC tests prove exactly one permitted fallback, no duplicate committed transactional write, no fallback for excluded modes, and accurate partial-failure reporting.
- [ ] Update generated/user-facing help and docs/spanner-driver-compatibility.md for the final behavior.
- [ ] Capture test output to files and check exit codes. Start with focused unit/fake-RPC tests, then run `make check` before any push. Run `make check-race` when changing transaction concurrency/lifecycle. Keep the coverage floor at 80%.

Keep each implementation PR confined to this behavior. Do not migrate to database/sql, copy driver infrastructure, or add unrelated dependency upgrades. Design-first tickets must resolve their stated contract before implementation.

## Accepted bounded implementation contract (2026-09-15)

The design uses the strong mutation-limit classifier in pinned [TransactionMutationLimitExceededException.java](https://github.com/googleapis/google-cloud-java/blob/dd2c41a0e2ec05d39b93f7d94169d596bb7a5483/java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransactionMutationLimitExceededException.java). Go Spanner v1.95.0 preserves an underlying gax `apierror.APIError` through Unwrap, while its GRPCStatus reconstructs a status without Help details. A local actual-SDK fake-RPC design probe reported retained Help after ExecuteStreamingSql (one SQL, zero Commit, one Rollback) and ExecuteSql; this is local evidence, not a managed-server wording guarantee. Implementation acceptance must reproduce this through the real mycli execution path. This section supersedes the design-first status for the narrow scope below.

### Setting and eligibility

- Add `AUTOCOMMIT_DML_MODE=TRANSACTIONAL_WITH_FALLBACK_TO_PARTITIONED_NON_ATOMIC`. Default remains TRANSACTIONAL. Existing TRANSACTIONAL and PARTITIONED_NON_ATOMIC behavior must remain unchanged.
- Only a single ordinary implicit/autocommit UPDATE or DELETE without a returning clause is eligible. No INSERT, explicit/pending owner, RO transaction, manual or automatic batch, SAVEPOINT/recovery owner, EXPLAIN/analysis-only path or other command. Capture eligibility before creating the implicit owner, not after failure retirement makes the session look idle.
- The initial transactional callback must have failed at its user ExecuteSql/ExecuteStreamingSql operation before any Commit was attempted. Distinguish that phase explicitly in the existing execution path. Never infer it from the final error code/string alone, never use a generic RunInNewOrExistRwTx error as proof, and never intercept Commit failure, even with otherwise matching Help metadata.

### Strong error predicate and single retry

- Require InvalidArgument, the complete sentence `The transaction contains too many mutations.`, and exactly one Help link with description `Cloud Spanner limits documentation.` and URL `https://cloud.google.com/spanner/docs/limits` on the underlying actual API error. Inspect retained typed details via errors.As/Unwrap. A hand-built spanner.Error with only its public Code/Desc is insufficient.
- Exclude the generic `Transaction resource limits exceeded` branch, message-only errors, missing/wrong/extra Help links, ResourceExhausted, Aborted, cancellations, deadlines and unknown/ambiguous transaction outcomes. Preserve the original failure when no strong predicate is established.
- Finish original-owner rollback/retirement through the existing failure path before starting PDML. Verify no active/recovering owner remains; never retain/reuse the failed physical transaction. Do not create a generic retry framework or add ABORTED retry (#293).
- Submit at most one fallback through the existing partitioned-update implementation. If PDML rejects unsupported/non-partitionable SQL, report that error together with the original cause and stop. Never fall back recursively or retry after partial PDML failure.

### Captured inputs, deadline and result

- Parse/freeze the statement and bound parameter values once. Freeze relevant priority, request tag, optimizer and supported query settings for the attempt; later retirement/LOCAL restoration must not cause re-parsing from live session values. Keep existing normal PDML behavior unchanged while reusing its concrete execution helper. Transaction-specific LastStatement/transaction tags must not leak into inappropriate PDML options; document any options the SDK does not support.
- Preserve one caller/STATEMENT_TIMEOUT budget across both phases. Capture the logical owner's absolute TRANSACTION_TIMEOUT deadline before its retirement, and apply the remaining deadline to fallback from the still-live original caller context. Do not reuse the owner context that is deliberately canceled on retirement or start a new duration. If either budget is exhausted, do not issue PDML.
- On success return the observed lower-bound PDML affected-row count and make the non-atomic fallback visible. Do not retain transactional commit stats/timestamp, fabricated exact count or stale query-plan/result-cache data from the failed attempt. On failure preserve both the initial error and the fallback error and warn that PDML can have partially committed; never claim rollback of that phase.
- Explain the opt-in's atomicity change, eligibility limits and pinned classifier in generated help, user docs and compatibility matrix. No stable/Java-complete parity claim.

### Acceptance tests and delivery

- Actual fake Data RPCs through Session/statement execution must verify eligible SQL failure -> Rollback -> one PDML, Commit=0 for the original; positive retained Help through pinned SDK; and a matching Commit-phase failure with zero PDML.
- Cover every negative eligibility/classifier case, returning DML, pending/explicit owner, batch/SAVEPOINT, failed PDML and partial-failure message. Check wrapped causes, exact RPC sequence/counts and no recursive fallback.
- Capture actual transactional/PDML request parameters and relevant options to prove identical frozen inputs. Exercise cancellation/expired budgets between failure and fallback, remaining transaction deadline across retirement, and no restart of statement timeout. Use event/context synchronization rather than fixed waits.
- Keep default and explicit PARTITIONED_NON_ATOMIC regression coverage; update enum generated outputs and docs with repository tools. No dependency upgrades or telemetry/DDL-policy implementation in this PR.
- Capture focused tests and full make check to files and check exit codes. Full make check and make check-race must pass before every push; use actual Docker where required. One Issue/PR on an isolated branch, additive main integration only, repository Apache2.0 headers, coverage floor80 unchanged.

apstndb:
## Summary

Adds opt-in `AUTOCOMMIT_DML_MODE=TRANSACTIONAL_WITH_FALLBACK_TO_PARTITIONED_NON_ATOMIC` for issue #964. Default remains `TRANSACTIONAL`. Existing `TRANSACTIONAL` and `PARTITIONED_NON_ATOMIC` paths are unchanged.

After one ordinary implicit `UPDATE`/`DELETE` fails at user `ExecuteSql`/`ExecuteStreamingSql` **before any Commit**, and the unwrapped API error is `InvalidArgument` containing `The transaction contains too many mutations.` plus exactly one Help link (`Cloud Spanner limits documentation.` / `https://cloud.google.com/spanner/docs/limits`), mycli rolls back/retires the original owner and submits **one** partitioned DML through the existing helper.

Eligibility is captured **before** the implicit owner is created. SQL, bound parameters, priority, request tag, and optimizer settings are frozen so later retirement or `SET LOCAL` restore cannot re-read live session values. Transaction tags and `LastStatement` are not sent on the partitioned attempt.

## Timeouts and results

- One caller/`STATEMENT_TIMEOUT` budget covers both phases. The 24h partitioned-DML default is not re-applied.
- The logical owner's absolute `TRANSACTION_TIMEOUT` deadline is captured before retirement and applied to fallback from the still-live caller context. Exhausted statement or transaction budgets skip PDML.
- Success returns the observed lower-bound PDML count and marks the result as a non-atomic mutation-limit fallback. Transactional commit stats/timestamp and stale query-plan cache from the failed attempt are not retained.
- PDML failure wraps both causes and warns that partitioned DML may have partially committed; that phase is not rolled back.

## Out of scope

INSERT, `THEN RETURN`, explicit/pending/RO/SAVEPOINT owners, manual/automatic batch, EXPLAIN/analysis, Commit-phase failures (even with matching Help), `ResourceExhausted`/`Aborted`/cancel/deadline, the weaker Java `Transaction resource limits exceeded` branch, #402 DDL policy, and #293 ABORTED retry.

This is not a Java-complete or stable driver-parity claim.

Fixes #964

## Gates

- Focused Session/fake-RPC tests: positive SQL-phase fallback, Commit-phase negative, classifier/eligibility matrix, frozen inputs, cancel and remaining-deadline budgets
- `make check` (actual Docker) and `make check-race` before push
- Coverage 86.9% (floor 80)

<div><a href="https://cursor.com/agents/bc-bae5a297-62a2-5f7c-8de5-bdef2c15733d?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-bae5a297-62a2-5f7c-8de5-bdef2c15733d&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


apstndb:
Independent review of ea42a61c6e91adb23d17bab76843ec59a28563fb: the scoped SQL/Commit phase separation, frozen inputs, rollback/retirement and one PDML call look consistent with #964. Focused actual-SDK tests and full Docker-backed `make check` passed locally (exit 0). Two test corrections remain before acceptance:

1. **[P2] Exercise Help negatives through the retained SDK/API error.** `internal/mycli/dml_mutation_limit_test.go:81-100` passes bare status errors for wrong URL/extra Help/weaker message. `isMutationLimitExceeded` requires an underlying `*apierror.APIError`, so these can reject before checking the intended predicate. Runtime mutation probe: replacing only the final Help description/URL comparison with `return true` still lets `TestIsMutationLimitExceededClassifier` pass. Please construct authentic APIError/Spanner wrappers or use the fake Data RPC path for wrong URL, wrong description, extra links, weaker text and wrong codes, with a corresponding strong positive. The named negative must reach its intended gate; this guard determines whether atomicity may be weakened.

2. **[P2] Remove the fixed future date from the deadline-phase test.** `internal/mycli/dml_mutation_limit_rpc_test.go:426` fixes the synthetic clock at 2026-09-15 12:00 UTC, but the resulting absolute deadline is passed to real `context.WithDeadline`/SDK calls. Once real time passes it, the test fails before the SQL-to-fallback boundary. An overlay changing only the date to an already elapsed year reproduced `DeadlineExceeded ... during getting session`, rather than the expected between-phase expiry. Anchor the fake clock to an invocation-relative real instant (or inject a consistent clock/timer boundary) and explicitly assert the between-phase hook ran after the expected original SQL/rollback and before zero PDML. Keep channel/clock synchronization instead of sleeps.

Please make an additive test fix, run the focused cases plus full `make check` and `make check-race` before push, then own applicable exact-head CI through green. Preserve branch history and main integration; no force-push or merge. Do not expand this into a retry framework or unrelated cleanup.

apstndb:
Both findings are addressed by fd3ee5437a7f5cea126a741b9f932bc6cfb645e2. Independent re-review of that exact HEAD found no remaining actionable findings in this PR's scope.

The new authentic APIError negatives now reject wrong Help URL/description; the previously passing mutation that removes those comparisons now fails at those two named cases. The deadline test uses invocation-relative time and verifies the actual SQL/rollback-to-fallback boundary. Focused actual-SDK tests passed locally. The production tree is unchanged from ea42a61, where the independent full Docker-backed `make check` passed; the implementer reports full `make check` and `make check-race` passed again on fd3ee54, and current applicable CI is green. No unresolved review threads or pending reviews were found.

The feature remains explicitly opt-in and limited to the documented SQL-phase mutation-limit predicate; this is not a general retry or managed-server wording guarantee.

PR #985, opened 2026-09-14
feat: add CLI_DDL_IN_TRANSACTION_MODE policy

apstndb:
## Summary

Implements `CLI_DDL_IN_TRANSACTION_MODE` (`FAIL` default, `ALLOW_IN_EMPTY_TRANSACTION`, `AUTO_COMMIT_TRANSACTION`) per the accepted #402 contract (2026-09-15). Spanner DDL stays non-transactional.

- Capture the policy on each logical owner, including pending `BEGIN`. Session `SET` after `BEGIN` applies to the next owner. `SET LOCAL` may change this owner only before user work; a later `SET LOCAL` is rejected without changing the registry.
- Empty means no admitted user database work, buffered mutations, queued DML, or SAVEPOINT history. Constructor `BeginTransaction`, `TRANSACTION_TIMEOUT` first-use, and heartbeat alone do not count.
- Pending empty: `FAIL` rejects; `ALLOW`/`AUTO_COMMIT` retire without constructing or committing.
- Constructor-only empty RW: `ALLOW` rolls back; `AUTO_COMMIT` commits; then DDL.
- Nonempty RW: only `AUTO_COMMIT` flushes eligible automatic DML, commits, then runs DDL. Manual DML batch is never implicitly `RUN`/`ABORT`.
- RO and SAVEPOINT recovery reject with zero Admin RPCs.
- `DdlStatement`, `BulkDdlStatement`, and `SYNC PROTO BUNDLE` are non-transactional mutations so pending `BEGIN` is not turned into constructor `BeginTransaction` before admission. `CreateDatabase` stays out of scope.
- Descriptor serialization is validated before irreversible commit/retirement. Empty BulkDdl is a no-op and does not commit. `START BATCH DDL` is admitted before batch state changes; `RUN BATCH` rechecks.
- After a successful commit, a later DDL failure reports both the commit and the DDL cause. EOF/`EXIT`/`Close` never auto-commit because of this variable.
- Does not implement AUTOCOMMIT (#83), default-sequence repair (#963; owned by #984), or generic retry.

Fixes #402

## Test plan

- Focused policy/fake Data+Admin RPC tests for the admission table, SET LOCAL, START/RUN, empty BulkDdl, commit-then-failed DDL, automatic count mismatch, RO/recovery, SAVEPOINT history, and all three `DDL_EXECUTION_MODE`s.
- Emulator proof that data committed before a later failed DDL remains committed and cannot be undone by a later `ROLLBACK`.
[...]

Issue #83, opened 2024-12-09
Implement AUTOCOMMIT using the existing logical transaction lifecycle


apstndb:
## Current scope

Implement `AUTOCOMMIT` through the existing logical transaction lifecycle. Default `true` preserves today's single-use reads and implicit RW DML commits; `false` lazily starts `BeginPendingTransaction` at the next eligible database operation. Explicit BEGIN remains available. No second transaction engine or automatic whole-transaction retry.

## Accepted implementation contract

- Register in `varDefs`, set the actual Go default to `true`, and remove the unimplemented placeholder. Use the existing validation/error conventions with a small custom setter where needed.
- Reject value changes during pending/active/recovering logical ownership or an open manual batch. Same-value SET is idempotent without replacing the owner; RESET follows the same rule. No SET LOCAL. Changing false to true never commits anything implicitly.
- After explicit COMMIT/ROLLBACK, remain idle until the next eligible operation. Do not eagerly install a replacement that hides LOCAL cleanup errors. Preserve explicit BEGIN priority/isolation and existing READONLY capture when pending ownership is determined.
- Centralize eligibility separately from `MutationStatement`. Eligible operations: ordinary SELECT, DML including THEN RETURN, buffered MUTATE, EXPLAIN ANALYZE for queries/DML, automatic DML enqueue, RUN BATCH DML, and the first enabled SAVEPOINT. Required admission and local input validation still precede irreversible effects; a disabled or invalid SAVEPOINT must not acquire a new owner.
- HELP/SET/SHOW VARIABLES/SHOW TRANSACTION, metadata/admin SHOW, EXPLAIN PLAN, explicit transaction commands, START/ABORT BATCH, DDL (including proto DDL), Create/Drop database, ADD SPLIT POINTS, TRUNCATE/explicit PDML, partition commands, EXPORT/DUMP, BQ/CQL/LLM and USE/DETACH do not auto-begin an owner. Preserve their existing state-specific guards. EXPLAIN PLAN DML while idle remains one-shot and does not leave a transaction open; it may use an existing explicit owner according to current behavior.
- Ordinary DML with AUTOCOMMIT=false must use its logical owner; AUTOCOMMIT_DML_MODE and mutation-limit PDML fallback cannot silently bypass transaction grouping. DDL uses the merged #402 policy; do not add a separate hidden commit or recreate ownership after DDL.
[...]

PR #989, opened 2026-09-14
feat: expire idle transactions independently of keepalive

### LINKED ISSUE #357: Expire user-idle transactions independently of keepalive traffic

apstndb:
Fixes #357.

Implements `CLI_IDLE_TRANSACTION_TIMEOUT` and `--idle-transaction-timeout` from the live Issue 357 contract (2026-09-15 body). This is a sliding user-idle quiet interval on the logical owner, independent of `KEEP_TRANSACTION_ALIVE` and distinct from `TRANSACTION_TIMEOUT` (#482).

## Behavior

- Duration strings (`60s`, `5m`) or `NULL`. `NULL`/`0` stays disabled. No default 60s. Negative and overflow values are rejected.
- Capture at `BEGIN`. Session `SET`/`RESET` after `BEGIN` applies to the next owner. `SET LOCAL` is allowed before admitted user work, then freezes.
- Successful explicit `BEGIN RW` / `BEGIN RO` that acquired a server transaction starts the first quiet interval. A resource-free pending `BEGIN` does not arm.
- Completed admitted work rearms, including successful buffered DML/`MUTATE` and successful `SAVEPOINT` / `RELEASE` / `ROLLBACK TO`.
- Heartbeat `SELECT 1`, client-only `SHOW`/`SET`/`RESET`, invalid syntax, and rejected admission do not reset idle.
- Expiry holds through the user RPC, iterator consumption, and CLI result rendering/pager. Idle is not an RPC context deadline.
- Expiry rolls back and retires only the matching owner. The next ordinary command reports a one-shot error without executing. `ROLLBACK`/`CLOSE`/`BEGIN`/`USE`/`DETACH` acknowledge the notice.
- After #985, `idleUserWork` remains a separate bit from `CLI_DDL_IN_TRANSACTION_MODE` `hasUserWork`: constructor `BEGIN RW` arms idle but does not mark #402 user work.

## Checks

- `make check` exit 0
- `make check-race` exit 0
- Logs under `/workspace/.tmp/driver-parity-357/`

Draft until Codex accepts. No rebase/force-push. Unpublished `cursor/idle-transaction-357-design-72f8` and published `cursor/sequence-kind-963-72f8` are preserved.

<div><a href="https://cursor.com/agents/bc-b73f6b55-6bbd-507b-ac6b-7179a6c972f8?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-b73f6b55-6bbd-507b-ac6b-7179a6c972f8&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


apstndb:
Reviewed `2701f41150d1987edb54cbe8256f4fb4ec125952` against main `34dba40fae5870dccf288521a852d11bac4e56a4`. Three accepted-contract regressions are reproduced with the existing real SDK/fake-RPC harness and actual Session/CLI dispatch, using the clock seam and expiry callback rather than sleeps:

1. **[P2, runtime-reproduced] A client-only SHOW crossing the old deadline restarts idle.** `idle_transaction.go:214-215` records `idleNeedsRearm` for a held deadline; `:179-191` later schedules `now + idle` without consulting the prior quiet interval or whether user work occurred. In an armed owner, I ran `Cli.executeStatement` for `SHOW VARIABLE CLI_IDLE_TRANSACTION_TIMEOUT`, advanced the test clock past the deadline and fired the watcher inside the result writer. After output returned the owner received another full interval. Repeating inspection around deadlines can keep an otherwise idle transaction alive. Preserve the existing deadline for non-activity commands and retire at the safe completion barrier if it has elapsed; grant a fresh quiet interval only for completed admitted work.

2. **[P2, runtime-reproduced] An owner created by the current CLI command is not held through its result.** `Cli.executeStatement` calls `beginIdleResultHold` before dispatch, but `idle_transaction.go:104-105` does nothing if there is no owner yet. `BEGIN RW` then creates and arms an owner; firing expiry from its actual result writer retires that owner before CLI output completes. The regression fails with "BEGIN RW owner expired while CLI result output was still running". Hold the CLI execution/result frame so a newly installed owner is covered too, while retaining owner-safe release. Also inspect the error-unwind order: the current hold defer runs before the earlier `sink.abort` defer, so pager/error cleanup must remain inside the hold as well (this latter ordering observation is source-backed, not separately runtime-reproduced).

3. **[P2, runtime-reproduced] Successfully buffered manual DML does not count as activity.** The new hooks cover automatic enqueue, but `bufferOrExecuteDML` at `execute_dml.go:46-56` appends manual DML and returns without an idle activity hook. With an already armed owner, actual `START BATCH DML` followed by `INSERT INTO t (id) VALUES (1)` leaves the idle generation unchanged. A user building a manual batch can therefore lose the transaction despite admitted work. Rearm only after successful manual enqueue, preserving no-rearm for rejected input/THEN RETURN/admission errors.

The combined overlay exits 1 with those three specific assertions; source and fixture compilation completed. Existing idle tests manually acquire holds or fire a second deadline after releasing them, so they miss the actual CLI and expired-SHOW cases. Please add real dispatch regressions for these boundaries, preserve total-timeout precedence/LOCAL safe restoration/stale-owner protection and simplify unused state if the corrected lifecycle makes it unnecessary. The file header's claim that #402 is not on this tree is also stale; keep the distinct constructor-vs-DDL admitted-work rationale but update that comment.

Make one additive fix and run focused lifecycle tests, full `make check`, and `make check-race` before pushing. This lane retains routine CI ownership; return fixing HEAD and exits. No change to the disabled default, 80% coverage floor or publication authority. #83 implementation is running separately and will integrate these owner/result hooks after acceptance.

apstndb:
Re-reviewed `dd2d0c4028c10126866b7fcbe4ead3f8b2bff014`. All three original independent regressions now pass, along with the focused idle suite. The manager-level CLI hold, preserved elapsed deadline and manual DML activity hook address the earlier findings. Two adjacent lifecycle gaps remain:

1. **[P2, runtime-reproduced] Final CLI idle retirement leaves SET LOCAL active until another command.** `idle_transaction.go:111-123` can now retire the owner when the final result hold is released, after Session's restoration defers already ran, but does not drain the detached undo. In an actual CLI SHOW writer, I fired expiry while `SET LOCAL CLI_VERBOSE=TRUE` was active over a FALSE session value. The owner remained alive and LOCAL remained TRUE during output (correct); after `Cli.executeStatement` returned the owner was gone but LOCAL was still TRUE and one pending undo entry remained. The next prompt/line-editor configuration can therefore observe expired transaction-local settings. Restore detached undo outside the lock at this final serialized CLI completion boundary, after pager/error cleanup, while preserving nested-frame/owner safety and never invoking registry setters from a timer. Cover both active output and final restoration; do not hide this by issuing another statement in the assertion.

2. **[P2, runtime-reproduced] PROFILE SELECT activity depends on SAVEPOINT capture being enabled.** `statements_explain_describe.go:472-479` calls `finishQueryCapture`, whose activity notification at `savepoint_capture.go:202-204` runs only when a capture token belongs to the owner. With default SAVEPOINT support disabled, actual `EXPLAIN ANALYZE SELECT 1` and `SET CLI_QUERY_MODE='PROFILE'; SELECT 1` drain the RPC but leave the idle generation unchanged. Both independent probes fail; the fake server returns the precise expected missing-plan emulator error after the query ran. Ordinary SQL has a separate no-token activity hook in `execute_sql.go`, but the analysis path does not. Count completed admitted query work consistently regardless of whether a replay token exists, including a surviving-owner server/result error. Keep rejected local admission and client-only SHOW/SET from becoming activity. Prefer a small shared completion hook over another divergent copied lifecycle path.

These are additional actual-dispatch tests, not a request to rewrite the transaction engine. Keep the now-passing original regressions. Please fix additively, incorporate current main `224d3053290c4d2d28ee1516e41f8571a243a239` if not already underway, and run focused tests, full `make check` and `make check-race` before the next push. This lane continues to own routine CI. Return the exact fixing HEAD and command exits; final acceptance remains pending. No API duplicate of this GitHub feedback is needed.

apstndb:
Reviewed HEAD: b1665f56b0fa847c84f00ef46601ab54adeb3dd6. The five earlier independent regressions now pass, and independent `make check` passed at 381a1639105afda05cf1eae99ffd7e34f51c450b before the additive #990 integration. One additional owner-safety regression remains in the new common completion hook.

**[P2, runtime-reproduced at the transaction-manager boundary] Do not count stale query completion as activity of a replacement owner.** `internal/mycli/idle_transaction.go:140-142` calls `finishQueryCaptureLocked(tok, ...)`, which correctly ignores a stale token, then unconditionally updates whichever `tm.tc` now exists. The prior `finishQueryCapture` deliberately checked `tok.belongsToLocked(tm)` before updating activity. This loses both logical-owner and physical-attempt protection.

Independent reproduction on b1665f5: enable capture and idle timeout using the existing heartbeat harness; admit `SELECT 1` through `runQueryWithStatsAndCapture`; drain the actual iterator with `consumeRowIterObserving` and finish its receipt; call the real manager ROLLBACK and BEGIN APIs to replace the logical owner; then deliver the old `completeAdmittedQuery(tok, nil)`. The replacement owner's idle generation advances from 1 to 2. This is a manager-level delayed-completion reproduction, not a claim that the normal serial CLI creates concurrent commands. It regresses the stale-completion contract already exercised by `TestSavepointOwnerCaptureTokenIgnoresUnadmittedAndStaleCompletion` and matters to the shared owner lifecycle.

Please retain capture-token owner/attempt validation before recording activity, while preserving the just-fixed normal/PROFILE completion behavior when SAVEPOINT capture is disabled. Add a focused delayed old-owner/attempt completion regression alongside a positive no-capture query completion test. Preserve the original error and pending/reservation state of the replacement; do not introduce a broad observer framework. The previous five fixes should stay intact.

Local reproduction command: `GOTOOLCHAIN=go1.26.8 go test -short -count=1 -overlay .tmp/idle-timeout-review/stale-review-overlay.json ./internal/mycli -run '^TestReview989StaleQueryCompletionDoesNotRearmReplacement$'` exited 1 at the intended generation assertion after actual iterator consumption. Before the next additive push, run the focused suite, `make check` and `make check-race`, capture logs and check exits, and let the existing coordinator handle routine CI. Keep the PR draft until exact-head acceptance.

apstndb:
ee743534b1a940e0b5f0ff57d62a4a7a88250912

apstndb:
Independently reviewed current HEAD **ee743534b1a940e0b5f0ff57d62a4a7a88250912** against current main **0ffacc862fe304e8db78c41a2f2912a31bc36793**. **No actionable findings remain in the reviewed scope.**

The final fix preserves the logical-owner/physical-attempt check before recording query activity and retains normal/PROFILE completion when SAVEPOINT capture is disabled. All six independent regression probes now pass, including the actual-consumption stale-completion reproduction; the focused idle and existing stale-capture tests also pass. Independent full `make check` passed on this exact HEAD (tests, lint and formatting). The worker reports focused/full/race gates passed before its push.

Current GitHub checks all pass, unresolved review threads are empty, and the branch contains the current main with clean merge state. The prior final-CLI LOCAL restoration, SHOW/deadline handling, newly created owner output hold and manual DML-buffer activity fixes remain intact. The manager-level delayed-completion test is scoped evidence; this review does not claim normal serial CLI commands run concurrently. Ready for the authorized squash merge.

Issue #725, opened 2026-07-05
System-variable schema metadata model (varDef): declarative defs feeding registry, flags, guards, RESET ALL, SET LOCAL, and docs


apstndb:
# Design: System-Variable Schema Metadata Model (`varDef`)

Status: design (no code changed yet). Citations are against main at b5057e5; paths are repo-relative under `internal/mycli/` unless noted.

## 0. Current state, verified

The registry today is a runtime map built imperatively:

- `VarRegistry.registerAll()` (var_registry.go:162-516) makes ~87 `r.Register(name, handler)` calls. Metadata per variable is smeared across the handler: description and read-only flag live *inside* each handler (`VarHandler[T].description/readOnly`, var_handler.go:40-48), guards are hand-rolled `CustomVar` setters (READONLY's in-transaction check at var_registry.go:166-180; CLI_ENABLE_ADC_PLUS's session-init-only check at var_registry.go:486-499, with the acknowledged TODO at var_handler.go:33-39).
- `Variable` is `{Get, Set, Description, IsReadOnly}` (var_handler.go:12-17); 10 handler types each re-implement `Description()`/`IsReadOnly()` boilerplate (var_custom_handlers.go, var_enum_handlers.go).
- COMMIT_RESPONSE and CLI_DIRECT_READ are special-cased at **6 sites**: SHOW VARIABLES (statements_system_variable.go:51-66), SET LOCAL (statements_system_variable.go:116-118), HELP VARIABLES (statements_system_variable.go:202-215), setFromGoogleSQL/setFromSimple (system_variables_registry.go:30-38, 65-73), and get (system_variables_registry.go:137-165), plus the tombstone comment at var_registry.go:515.
- Flag wiring is a dual write path: `createSystemVariablesFromOptions` (config.go:342-456) direct-assigns ~25 fields, bypassing setters. Concrete divergences: `--prompt2=""` is accepted (config.go:369-371) though the registry setter rejects empty (var_registry.go:244-253); `--log-level` is parsed by `SetLogLevel` (config.go:348) duplicating `LogLevelVar.Set` (var_custom_handlers.go:257-273); STATEMENT_TIMEOUT's default lives only in a kong tag (`default:"10m"`, config.go:159) while RPC_PRIORITY's default lives in *both* `newSystemVariablesWithDefaults` (system_variables.go:341) *and* the mapping row `cmp.Or(opts.Priority, "MEDIUM")` (config.go:591).
[...]

Issue #292, opened 2025-06-17
Enhance system variables documentation with comprehensive user guide


apstndb:
## Summary

The `docs/system_variables.md` file needs comprehensive enhancement to provide a complete user guide for system variables and related functionality in spanner-mycli.

## Current State

- File contains only TODO placeholders
- No documentation for 60+ system variables available to users
- Missing coverage of system variable features and usage patterns

### 1. Complete System Variables Documentation

**Spanner JDBC Compatible Variables**
- Document variables that mirror Spanner JDBC functionality
- Include: AUTO_PARTITION_MODE, AUTO_BATCH_DML, AUTOCOMMIT_DML_MODE, OPTIMIZER_VERSION, etc.

**CLI-Specific Variables**  
- Document CLI_ prefixed variables for spanner-mycli specific functionality
- Include: CLI_FORMAT, CLI_AUTOWRAP, CLI_ENABLE_HIGHLIGHT, CLI_EXPLAIN_FORMAT, etc.

### 2. System Variable Features Documentation

**Variable Management Commands**
- `SHOW VARIABLES` - Display all system variables
- `SET variable_name = value` - Set variable values
- Variable querying and inspection

**Variable Categories**
- Connection settings (CLI_PROJECT, CLI_INSTANCE, CLI_DATABASE)
- Display formatting (CLI_FORMAT, CLI_AUTOWRAP, CLI_FIXED_WIDTH)
- Query execution (AUTOCOMMIT_DML_MODE, RPC_PRIORITY, STATEMENT_TIMEOUT)
- Output control (CLI_USE_PAGER, CLI_ENABLE_HIGHLIGHT, CLI_MARKDOWN_CODEBLOCK)
- Development tools (CLI_VERBOSE, CLI_LOG_LEVEL, CLI_LINT_PLAN)

### 3. Detailed Variable Reference

For each variable include:
- **Name and type**
- **Default value**
- **Description and purpose**
- **Valid values/range**
- **Usage examples**
- **Related variables**

### 4. Usage Examples and Patterns

**Common Use Cases**
```sql
-- Display configuration
SET CLI_FORMAT = 'VERTICAL';
SET CLI_AUTOWRAP = true;

-- Query optimization
SET OPTIMIZER_VERSION = 'LATEST';
SET RPC_PRIORITY = 'PRIORITY_HIGH';

-- Development workflow
SET CLI_VERBOSE = true;
SET CLI_ECHO_EXECUTED_DDL = true;
```

[...]

PR #995, opened 2026-09-14
docs: clarify system-variable precedence, defaults, and links

### LINKED ISSUE #993: Clarify system-variable precedence, defaults and documentation links
## Problem and scope

The generated system-variable inventory already exists (#710), but the surrounding user guide still has unclear startup/runtime precedence and misleading defaults. README currently gives `CLI_DATABASE_DIALECT = "TRUE"` as a default, omits `--set` from the precedence explanation, and does not clearly distinguish a stored NULL timeout from its execution-time fallback. Related query-plan documentation needs direct cross-links.

This is the first bounded documentation phase of #292, following the current-source #725 PR6 audit. Keep #292 open for its remaining richer generated metadata/guide scope. Do not repeat the completed registry generator or the intentionally deferred #725 startup-table relocation.

## Implementation

- Update `README.md`, `docs/system_variables.md` and `docs/query_plan.md` using current source as the authority. Keep prose outside the generated sysvars marker block.
- Explain built-in defaults, home/current-directory TOML, environment, flags, then `--set`; describe `--init-command` / `--init-command-add` as subsequent SQL and preserve the existing RESET startup-snapshot boundary. Verify actual special/presence-dependent options rather than asserting every variable supports every source.
- Distinguish registered/stored value, startup snapshot restored by RESET, current session value, captured transaction policy and execution-time fallback. Use `STATEMENT_TIMEOUT` as the concrete NULL versus ordinary/partitioned execution fallback example; do not introduce behavior changes.
- Correct the invalid database-dialect default/example and other concrete parser-invalid nearby examples. Use actual values from the existing registry and docs. Avoid claiming all JDBC-inspired variables have identical driver semantics.
- Add focused bidirectional links for query-plan columns/inline stats/templates and configuration guidance, reusing existing detailed feature docs rather than copying them. Remove obsolete placeholder framing only where the existing generated inventory or linked guidance actually covers the topic.
- Preserve the existing three-column HELP/reference output and generator. Do not add typ/defaultDisplay/examples/related fields, a new doc framework, a hand-maintained duplicate variable inventory, or runtime feature changes in this PR.

## Acceptance

- User-visible examples parse and refer to current supported names/values.
- Precedence and RESET/LOCAL explanations match the current implementation, including ordinary SET after SET LOCAL and excluded RESET surfaces.
- Actual stored NULL versus effective timeout behavior is explicit; runtime/session-dependent values are not presented as universal constants.
- Links point to existing headings; generated blocks remain reproducible.
- `make docs-update` produces no unrelated manual changes. Use an existing doc/link validation mechanism if available. Do not add a new test that merely asserts prose headings or mirrors static wording; validate examples with existing parser helpers or a bounded scratch proof.

## Verification and delivery

Capture output to a file and check each command's exit code. Verify relevant examples/link targets, run `make docs-update`, then repository-required `make check` and `make check-race` before every push. Preserve generated reference determinism and the 80% coverage floor. One feature branch and one PR fixing this child Issue, with `Part of #292` and a reference to #725 PR6; do not close either umbrella. No dependency upgrades, changes to defaults/setters, or edits to `.gemini/styleguide.md`.

Baseline audited: main `0ffacc862fe304e8db78c41a2f2912a31bc36793`. Integrate later main additively and refresh text affected by newly merged variables.

apstndb:
The system-variable guide now documents configuration precedence, startup RESET snapshots, and the difference between stored values and execution fallbacks. It replaces the duplicated README variable inventory with links to the generated reference, corrects invalid dialect/SHOW examples, and connects the query-plan documentation in both directions.

The change is prose-only. The existing three-column reference and runtime behavior are unchanged. Configured timeout values and supported pre-work SET LOCAL changes are described explicitly.

Validation: parser/config/RESET tests, local link checks, make docs-update, make check, make check-race, and all applicable CI passed. Documentation regeneration leaves no diff.

Fixes #993
Part of #292; relates to #725. These parent issues remain open.


apstndb:
Reviewed the complete documentation diff at b6f3eb897435e03d99c103d96cd9100c04c3ad0f against current main da9b07688c3e2125a45237e98469fa2329a605db. The inventory consolidation, parser examples, RESET startup ordering and stored/effective timeout distinction are useful. Three wording corrections are needed before acceptance:

1. README.md, Configuration Precedence: the new sentence says “later sources win”, but the list is highest-to-lowest. Either list in application order or say “in descending priority”.
2. docs/system_variables.md, Stored values table: “Frozen on the logical transaction at creation” is too strong. The owner captures an initial policy, but supported SET LOCAL settings can update it before admitted work and then freeze. For example, the current CLI_IDLE_TRANSACTION_TIMEOUT reference explicitly allows that change. Describe initial capture and future-owner session SET/RESET separately from the variable-specific pre-work SET LOCAL rules. Do not imply that every listed property is immutable immediately at BEGIN.
3. Qualify the repeated “omitting --timeout / --idle-transaction-timeout leaves NULL” wording as the default when neither the corresponding TOML/config value nor --set supplies a value. The new precedence section otherwise contradicts those absolute statements; an omitted command-line flag does not erase configured values.

These are source-backed documentation findings, not runtime failures. Keep the correction prose-only and concise. Please also describe the order in user-facing terms rather than implementation function names (newSystemVariablesWithDefaults / ApplyFlags). No new static prose tests are needed. Preserve the existing generated reference and other scopes.

Continue on this same branch with additive commits. Before any push, run relevant checks, make check and make check-race with captured logs and verified exits, and integrate current main if needed. Cursor owns routine CI; Codex owns exact-HEAD review and merge. No new agent or extra PR.

apstndb:
Re-reviewed the complete documentation change and the three corrections at 1c7927c300c07ddb65c577e49b3a0941e0992e8a. All findings from the earlier review are addressed by this commit; no actionable findings remain in the reviewed scope.

Independently checked configuration application order, RESET capture, supported owner-local changes and the corrected examples against source. make docs-update exited 0 and left no tracked diff; focused configuration/RESET tests passed. Runtime Go code, module files, Makefile and CI configuration are byte-identical to the previously locally validated ee74353 snapshot, so the full gate evidence is reused for unchanged code. The worker also reports make check and make check-race passing before this push. All applicable hosted CI passed at this HEAD.

This accepts the prose phase for #993; it does not complete the broader #292/#725 work.

PR #997, opened 2026-09-14
feat: retry ABORTED implicit read-write operations (#994)

### LINKED ISSUE #994: Retry ABORTED implicit read-write operations within a bounded logical lifetime
## Goal and dependencies

First implementation phase of #293: retry one implicit read-write operation when Spanner conclusively returns ABORTED, preserving its logical lifetime and publishing only the successful attempt's result. #293 remains open for explicit journal replay and observation completion in the next phase. Use current main containing AUTOCOMMIT (#990), idle expiry (#989), total timeout (#482), mutation-limit fallback (#964), and existing SAVEPOINT infrastructure.

## Property and staged scope

Implement `RETRY_ABORTS_INTERNALLY` as an opt-in boolean, default FALSE. Document that Java/Go reference drivers default TRUE and that this property controls application logical retry, not all SDK RPC/stream retries. Ordinary session SET/RESET affects future owner snapshots, SHOW displays the session value, and RESET restores the startup snapshot.

This first phase supports implicit RW only. With TRUE captured for a newly requested explicit/pending RW owner (including lazy AUTOCOMMIT=false), reject before beginning its database work with a clear explanation that explicit retry is not implemented yet. Do not silently ignore the requested policy. Do not reject read-only transactions or unrelated non-RW operations. An existing owner keeps its captured policy when ordinary SET changes the future-session policy. Reject SET LOCAL for this property in phase1; pending-owner LOCAL activation belongs to the explicit phase. Keep FALSE/default behavior and all other existing variables' semantics unchanged.

## Retry contract

- At most 50 attempts for the current implicit logical operation, including its first execution. Retry only `spanner.ErrCode(err) == codes.Aborted` from actual SQL, BatchDML or Commit. Never retry local validation/rendering errors or ambiguous Commit outcomes such as Unavailable/Unknown/DeadlineExceeded/Canceled.
- Preserve caller cancellation and the original transaction deadline through retry/backoff; never restart either budget. Use server RetryInfo when provided, otherwise bounded interruptible backoff consistent with the design's 1–32ms policy. Clamp waiting to the remaining budget and use an injectable deterministic wait seam for tests, not fixed sleeps.
- Reuse one logical owner with new physical attempts and frozen constructor options. Keep timeout/heartbeat/idle identity and SET LOCAL cleanup correct; restoration happens once at logical completion. Internal reconstruct/backoff is not new user activity and does not extend the idle deadline.
- Freeze SQL parameters, mutation/batch inputs and request options before attempt1, including the one-shot STATEMENT_TAG. Do not rebuild options from mutated session state or re-read external data for another attempt. Use current transaction helpers, not a second journal or database/sql migration. An implicit operation does not allocate the explicit replay journal.
- Inventory existing callback paths: ordinary DML, PROFILE/EXPLAIN ANALYZE DML, THEN RETURN, an implicit BatchDML operation and MUTATE. Preserve their existing eligibility and output semantics; do not accidentally include PLAN, partitioned DML or unrelated RO queries merely because helpers are shared. Current explicit/automatic-batch owners retain their default behavior until the explicit phase.
- Discard per-attempt row counts, returned body, metadata/plan/stats and cache changes after SQL succeeds but Commit aborts. Re-enter the operation with fresh output state, and publish one successful result only after Commit. A failed/current partially visible output is not retryable; prior attempts must not leak a PreparedBody or stale plan.
- Preserve #964 provenance: SQL and Commit remain separate phases. Exhausted ABORTED and every Commit failure are ineligible for mutation-limit PDML fallback. Do not label arbitrary non-ABORTED errors as "transaction was aborted".

## Target code and acceptance

Use `internal/mycli/transaction_manager.go`, current owner/deadline/reconstruct helpers, `execute_dml.go`, PROFILE DML helpers, batch/mutation callbacks, system-variable definitions and existing fake-RPC/receipt/timeout tests. Keep changes surgical and explain any additional file/API needed.

Verify actual callback/RPC paths for: SQL ABORTED then success; SQL success/Commit ABORTED then success; exhausted50; cancellation during backoff; original deadline exhaustion; unchanged non-ABORTED/ambiguous outcomes; frozen params/tag/options; same logical owner/new attempt; exactly-once cleanup and output/cache publication. Cover returned rows/PROFILE, implicit batch and mutation callbacks, and explicit/RO/PLAN/PDML exclusions with real error causes and call counts. Default FALSE must not gain hidden retries or journaling. Additive main integration must retain #989's stale-completion and CLI-hold behavior.

Update user docs and the driver-compatibility matrix with the opt-in default, implicit-only restriction, attempt bound and SDK distinction. Do not claim complete #293 parity or a stable contract. Capture focused test output and exits, run `make docs-update`, `make check` and `make check-race` before every push; coverage floor80. One feature PR fixing this child Issue, `Part of #293`; do not close #293. No dependency upgrades, managed-service writes or release changes.

apstndb:
Add opt-in `RETRY_ABORTS_INTERNALLY` for implicit autocommit read-write operations. When SQL, Batch DML or Commit returns ABORTED, retry the whole operation up to 50 total attempts using fresh physical transactions while preserving the logical owner, request inputs, statement tag and original caller/transaction deadline. Only the successful result is published.

The default remains FALSE. New explicit read-write owners with retry enabled and SET LOCAL for this property are rejected; existing owners retain their captured policy. PLAN, partitioned DML and read-only operations keep their existing behavior. Explicit journal replay and observation completion remain under #293.

Validation: focused RPC and independent regression tests, full make check, and all applicable CI including race passed at 463d5ec2b01c9e5b98c2426f677a8c3d19c2cd22. Tests cover abort phases, bounded attempts, cancellation, SDK retry delays, owner snapshots, frozen inputs, deadline causes and excluded paths.

Fixes #994
Part of #293


apstndb:
Review at badd937897ccfa9a98389e07d9a642155037966e, including the additive Commit-attempt retirement fix. Three remaining findings were reproduced with the real Go Spanner SDK against the existing local RPC harness (no managed service):

1. **[P2] Preserve an already-created pending owner's captured FALSE policy.** `BEGIN; SET RETRY_ABORTS_INTERNALLY = TRUE; SET TRANSACTION READ WRITE;` currently fails with `errRetryAbortsExplicitUnsupported`. The new guard in `BeginReadWriteTransaction` reads the current session value even when activating the existing pending owner. Ordinary session SET must affect future owners only. Use the pending owner's captured policy for activation; retain the pre-RPC rejection for newly created explicit/pending TRUE owners. Independent `TestReview997PendingOwnerKeepsCapturedFalse` failed through actual `execSQL`, not a direct registry-only assertion. Location: [transaction_manager.go:968](https://github.com/apstndb/spanner-mycli/blob/badd937897ccfa9a98389e07d9a642155037966e/internal/mycli/transaction_manager.go#L968).

2. **[P2] Extract RetryInfo from the SDK error rather than its reconstructed status.** A fake-server ExecuteStreamingSql ABORTED with `RetryInfo{RetryDelay: 1234ms}` reached `abortRetryWait` with **30ms**, the random fallback. Pinned `spanner.Error.GRPCStatus()` reconstructs status from code/description and drops details; the raw-status unit test bypasses that wrapper. The pinned public `spanner.ExtractRetryDelay(err)` already handles this and validates the duration. Use the supported helper or the documented unwrapped APIError path, and test SQL/Batch/Commit errors through the SDK. Location: [retry_aborts.go:90](https://github.com/apstndb/spanner-mycli/blob/badd937897ccfa9a98389e07d9a642155037966e/internal/mycli/retry_aborts.go#L90).

3. **[P2] Stop at an expired owner budget, including during backoff.** `remainingAbortWaitBudget` ignores an expired owner when the caller still has a positive budget, while `clampAbortRetryDelay(..., 0)` treats zero as unlimited. The actual wait is also bound only to the caller context. A deterministic logical-clock/RPC probe advanced `tm.nowFunc` past the original deadline immediately after the second physical Begin succeeded; it still passed a positive delay to the wait seam. Fix the distinction between no deadline and an exhausted deadline, bind backoff to the original owner/caller deadline and retain the TRANSACTION_TIMEOUT cause. A server RetryInfo can make this much worse after finding 2 is corrected. This is a reproduced control-flow boundary with a fake clock, not a wall-clock latency measurement. Locations: [retry_aborts.go:117](https://github.com/apstndb/spanner-mycli/blob/badd937897ccfa9a98389e07d9a642155037966e/internal/mycli/retry_aborts.go#L117), [transaction_manager.go:1701](https://github.com/apstndb/spanner-mycli/blob/badd937897ccfa9a98389e07d9a642155037966e/internal/mycli/transaction_manager.go#L1701).

All three independent regressions fail at the current reviewed HEAD; the additional READONLY + BEGIN RW probe correctly rejects before a Begin RPC and is not a finding.

Please tighten the PROFILE regression too: accepting arbitrary errors merely because two RPCs occurred and the owner retired does not verify successful PROFILE results. Assert the actual expected outcome/cause with a suitable fixture. Add current-operation deadline and changed-result-on-retry checks through the real path rather than only retry predicates. The rest of the semantic review is still open; these are not a clean bill of health for all 20 files.

Continue on the same #997 branch, no new agent or extra PR. Keep explicit replay #293 out of this phase. Capture focused checks plus make check and make check-race with actual exit codes before every additive push; integrate current main normally. Cursor owns routine CI and Codex retains exact-HEAD review, replies and merge.

apstndb:
Re-reviewed the corrections at d5bf29e87db39dbf8ec304755195fa63c76134eb. All three original independent regressions now pass. The public SDK retry-delay helper and the stronger PROFILE/result assertions address the earlier feedback. Two adjacent cases still need correction:

1. **[P2] Check the captured pending policy instead of bypassing the guard for every pending owner.** `SET READONLY = TRUE; SET RETRY_ABORTS_INTERNALLY = TRUE; BEGIN; SET TRANSACTION READ WRITE;` currently succeeds. READONLY legitimately permits the pending owner to exist, but explicitly activating it as RW now skips the guard solely because it is pending, even though its captured retry policy is TRUE. That admits the explicit RW configuration this phase promises to reject; the explicit path does not implement retry. Preserve activation of captured-FALSE owners after session SET TRUE, but reject captured-TRUE RW activation before the constructor RPC. This sequence fails the independent actual `execSQL`/SDK regression. Location: [transaction_manager.go:973](https://github.com/apstndb/spanner-mycli/blob/d5bf29e87db39dbf8ec304755195fa63c76134eb/internal/mycli/transaction_manager.go#L973).

2. **[P2] Preserve which deadline expired.** `abortWaitBudget` combines caller and owner deadlines, but both exhausted-budget branches call `stopImplicitAbortForDeadlineLocked`, which unconditionally adds `errTransactionTimeout`. With the default NULL TRANSACTION_TIMEOUT, an exhausted caller/statement deadline is therefore reported as `TRANSACTION_TIMEOUT exceeded: context deadline exceeded`. A deterministic logical-clock probe with only a caller deadline, advanced after the first SQL ABORTED response, reproduces this incorrect error classification through the SDK path. This is a logical-clock control-flow test, not a wall-clock measurement. Return the caller deadline cause when only that budget expired; add the TRANSACTION_TIMEOUT cause only when the owner budget expired. Keep the original caller cancellation behavior and the now-fixed owner deadline checks. Locations: [retry_aborts.go:251](https://github.com/apstndb/spanner-mycli/blob/d5bf29e87db39dbf8ec304755195fa63c76134eb/internal/mycli/retry_aborts.go#L251), [transaction_manager.go:1704](https://github.com/apstndb/spanner-mycli/blob/d5bf29e87db39dbf8ec304755195fa63c76134eb/internal/mycli/transaction_manager.go#L1704).

Continue on the same branch, integrate current main 4b55359db9a109b70c8e1f7480662af8a1b4ff7a additively, and capture focused checks plus `make check` and `make check-race` with checked exit codes before pushing. No broader replay implementation or additional PR is requested. The full semantic review remains open.

apstndb:
Correction to finding 1 in https://github.com/apstndb/spanner-mycli/pull/997#issuecomment-5670278202: the stated SQL sequence does **not** activate a read-write transaction. With READONLY=TRUE, `BeginStatement.Execute` starts RO immediately. `SetTransactionStatement.Execute` intentionally does nothing when the owner is not pending. My independent test asserted an error without verifying the owner mode or a new Begin RPC, so that claimed user-visible reproduction was invalid. I retract that finding; it must not block acceptance.

The new captured-TRUE guard at 5ed8796ae0836515ad51858a426183fcda67ed30 is a separate manager-level invariant, and its test explicitly constructs that pending owner after closing RO. It must not be described as proof of the original SQL scenario. No further production change is requested for that scenario. I am correcting the independent control to assert that RO remains RO and no additional Begin RPC occurs.

Finding 2 (caller deadline incorrectly labeled TRANSACTION_TIMEOUT with a NULL owner timeout) was valid and now passes its original independent regression. The original pending-FALSE, SDK RetryInfo and owner-deadline regressions also pass. Full semantic acceptance is still in progress.

apstndb:
Independent review completed for 463d5ec2b01c9e5b98c2426f677a8c3d19c2cd22 against current main db7c30bc089f0be5f64974b0770c1d97bc908103. No actionable findings remain within the opt-in implicit-operation scope of #994.

Reviewed the full production and test changes, including logical-owner snapshots, physical-attempt reconstruction, SQL/Batch/Commit boundaries, frozen request inputs, retry-delay extraction and deadline attribution. The independent pending-FALSE, SDK RetryInfo, owner-budget and caller-deadline regressions pass. The corrected SQL read-only control confirms RO remains RO with no additional Begin RPC; the earlier claimed RW activation was retracted in https://github.com/apstndb/spanner-mycli/pull/997#issuecomment-5670453630 and is not a remaining finding.

Validation at this exact HEAD: independent/focused retry tests passed (exit 0, 0.795s); full make check passed (exit 0, including lint and formatting). All nine applicable GitHub CI jobs, including race, passed; unresolved review threads and pending reviews are empty. Explicit transaction journal replay and observation completion remain future work under #293; this PR closes only #994.

PR #992, opened 2026-09-14
feat: add CLI_STRING_QUOTE_MODE NONE/AUTO/ALWAYS

apstndb:
Add `CLI_STRING_QUOTE_MODE = NONE | AUTO | ALWAYS` for human-readable table and vertical output. The default NONE preserves existing output; AUTO quotes empty or potentially ambiguous strings, and ALWAYS quotes every non-NULL STRING. Both quoting modes escape embedded quotes, backslashes and nonprintable characters, so SQL NULL, the string NULL and a string containing quotes around NULL remain distinguishable.

Apply the policy to typed query/SHOW/DML results and nested ARRAY/STRUCT values, including streaming output. SQL NULL styling, non-STRING formats, export formats, headers and preformatted plans remain unchanged. Document AUTO's exact heuristic and the display escape grammar.

Validation at dc47e438737829fda3d7c3b08377715ecfca1be1: independent SDK streaming and focused quote tests, full make check, and all applicable CI including race passed. Tests cover mode/reset/local behavior, nested and escaped values, default/export byte preservation and actual streaming processors/SQL execution.

Fixes #762

Issue #284, opened 2025-06-16
Implement Ellipsis (text truncation) for table output


apstndb:
## Accepted scope (2026-09-17)

Add opt-in end truncation to the existing table cell-fit path. This replaces the old multi-variable proposal, following the [maintainer scope reduction](https://github.com/apstndb/spanner-mycli/issues/284#issuecomment-4677745880). Width allocation and shared buffered/streaming rendering already exist; do not recreate them.

### User-visible behavior

- Add session boolean `CLI_ELLIPSIS`, default `FALSE`, through the normal SET / SET LOCAL / SHOW / RESET registry. No dedicated CLI flag is needed; existing `--set` works.
- `FALSE`: preserve current output byte-for-byte.
- `TRUE` with a finite constrained screen: end-truncate table headers/cells to the widths already allocated by `CLI_WIDTH_STRATEGY`, using ASCII `...` as the marker. Existing go-tabwrap handling reduces the marker to `.` / `..` when the allocated width is one/two cells.
- Preserve existing width-source rules: `CLI_AUTOWRAP=false` ignores fixed/TTY width; unknown terminal width remains unconstrained. In these unconstrained cases, ellipsis is a no-op and later streaming rows retain the existing lossless wrapping behavior. For batch output, explicitly enable AUTOWRAP and set `CLI_FIXED_WIDTH` to request truncation.
- Preserve explicit newlines by truncating each existing line independently. Verbose `Name\nSTRING` headers must retain both lines. Do not turn multiline content into a single truncated line.
- Preserve grapheme clusters, full-width character accounting, tab visualization, cell metadata and ANSI style containment. The marker must fit the allocated display width; no style may bleed into borders or following cells.
- Apply only to TABLE / TABLE_COMMENT / TABLE_DETAIL_COMMENT. Keep CSV, JSONL, SQL exports and other formats unchanged.

### Implementation locations

Current inspected source: `cc0650ab`; start implementation from current main. Later #1038 changes only AUTOCOMMIT tests.

1. Add the display setting in `DisplayVars` / `var_defs.go`, copy it through `toFormatConfig`, and add the corresponding `FormatConfig` field.
2. Select wrapping vs truncation in `format/streaming_table.go` (`wrapHeaders`, `wrapRow`) using the existing `constrainedScreen` / allocated widths. Leave `tw.WrapNone`, width strategies and minimum column width unchanged.
[...]

PR #998, opened 2026-09-14
fix: preserve calculated widths in streaming tables

### LINKED ISSUE #996: Preserve calculated column widths in streaming tables without truncating values
## Reproduced problem

With table streaming enabled, the current formatter silently truncates row values even when the preview has already seen the complete value and the screen has sufficient width. A one-column header `id`, preview row `abcdefghij`, screen width80 and preview50 render `abcdef`. No quote-mode setting is involved. `internal/mycli/format/streaming_table.go` is byte-identical between current main da9b07688c3e2125a45237e98469fa2329a605db and the reviewed #992 headb38a5b7b3d0b0cf82a2b294aabd8479acf3c9aa3, so this is a pre-existing renderer issue discovered while reviewing that feature.

It also affects TABLE_COMMENT/TABLE_DETAIL_COMMENT. Through a real SDK fake-RPC query, the STRING containing quotes around NULL should display `"\"NULL\""` under #992 AUTO, but the actual table stream only keeps its prefix. VERTICAL retains the complete value. #992's test named StreamingMatchesBuffered currently renders a PresentationBody twice and does not execute the streaming formatter, so it missed this.

## Scope and likely cause

The existing `InitFormat` computes `f.widths` from preview rows, but enables tablewriter streaming without passing those widths to tablewriter before Start/Header. Tablewriter establishes its own stream widths from the short header. Verify this against the pinned API/source; the current `tablewriter.Config.Widths` / `tw.CellWidth.PerColumn` API is available, and the older StreamConfig.Widths field is deprecated. Prefer a small correct connection of the existing widths, not another width calculator or renderer.

Fix this on current main in a separate PR from #992. Preserve buffered/default output, existing screen-width allocation/wrapping/ANSI handling and row ordering. Values wider than the selected stream width must follow existing wrapping policy rather than being silently cut by a second unconfigured layer. Respect skip-column-names, headers-only preview, previewed rows and rows arriving after preview. Keep TABLE_COMMENT and TABLE_DETAIL_COMMENT framing semantics intact. Do not implement #284's ellipsis feature, change exports, or upgrade dependencies.

## Acceptance

- An actual streaming TableStreamingFormatter initialized with header `id`, preview `[abcdefghij]`, screen80 preserves the full value. Compare with the buffered formatter under the same configuration.
- Exercise normal and comment table modes, nonzero and headers-only preview, skipped headers, and later rows wider than the initial sample, with existing expected wrapping semantics. Cover ANSI/NoWrap cases using existing helpers where they affect width propagation.
- Add a bounded real query/processor streaming regression, assert `alreadyDelivered()` so an accidentally buffered test cannot pass, and assert complete value content. The reproductions do not require a real database.
- Preserve none/default and existing formatter regressions. Avoid a new broad abstraction or tests that only mirror assignment to Config.Widths.

## Verification

Capture focused output and check exits, then run `make check` and `make check-race` before every push with additive current-main integration. One feature PR fixes this Issue; #992 then integrates it additively and re-runs its actual streaming tests. Coverage floor80. No managed-service writes, release or unrelated refactor.

Independent failing proofs at b38a5b7: TestReview992ExistingStreamingFormatterDoesNotTruncate (no new quoting code) and TestReview992ActualRPCStreaming (TABLE/comment modes fail, VERTICAL succeeds). Both failed at the intended full-value assertion, with no sleeps. Baseline source equality was checked against mainda9b076.

apstndb:
Streaming tables could discard value suffixes because tablewriter locked column widths from short headers instead of the widths already calculated from preview rows. Pass calculated widths through tablewriter's supported configuration before streaming starts.

For unknown or unlimited screen widths, use the same natural header/preview widths for both wrapping and tablewriter. Wider later rows wrap without losing characters or allocating an unbounded border. Buffered formatting and export formats retain their existing behavior.

Validation: formatter/preview-processor/actual SDK streaming tests, independent regressions for empty and short previews with wider later values, `make check`, `make check-race`, and all applicable current-head CI. Independent review and local `make check` passed at 41013aed0fc0f74e264129346cde9260d2180af1, including current main.

Fixes #996.


apstndb:
[P1] The unconstrained-screen branch still silently truncates later rows at 715101a04063ff47ac9a281e48d8b054da27feed.

`InitFormat` locks tablewriter to `naturalPreviewWidths`, but `f.widths` remains the effectively unlimited wrapping budget. Therefore a value wider than the header/preview is not wrapped to the locked cell width and tablewriter drops its suffix. This affects the normal unknown/non-TTY `math.MaxInt` screen path, including CLI_TABLE_PREVIEW_ROWS=0 and any wider value after the preview window.

Independent reproduction with the existing `renderStreamingTable` helper:

- screenWidth=math.MaxInt, header `id`, empty preview, later value `abcdefghij` -> only `ab` is printed.
- same screen, preview row `ab`, later value `abcdefghij` -> both rows print `ab`.

The assertion joins all emitted wrapped cells before checking the complete value, so this is lost data rather than expected wrapping. Both regressions fail at the reviewed HEAD. The existing unconstrained-screen test only covers a value already present in the preview, while the later-row tests use width80 and bypass the new branch.

Location: [streaming_table.go:98](https://github.com/apstndb/spanner-mycli/blob/715101a04063ff47ac9a281e48d8b054da27feed/internal/mycli/format/streaming_table.go#L98). Keep tablewriter widths and the actual row-wrapping widths consistent, without an unbounded border allocation. Cover unknown-width empty/short preview plus wider later rows, including a late NoWrap value, so the width fix cannot silently discard data. A bounded stream may wrap when later rows exceed its established widths; preserving data remains required. Do not add ellipsis/truncation policy or broaden #284.

Also integrate current main41138139b1957a6206124d60e3c1aff970d67f5a (#995 prose). Continue on this same #998 branch; #992 remains held until this prerequisite merges. Run focused real-path checks, make check and make check-race with captured outputs and checked exits before every additive push. Cursor owns ordinary CI; Codex retains review and merge. No new agent/PR needed.

apstndb:
Reviewed b78fd160b8770933cc8f4b66b33c9b06bc3ac90b. The reported unconstrained-screen data loss is fixed, and I found no further actionable issues in the current three-file diff.

The original independent empty/short-preview regressions now pass, together with the streaming formatter tests. Independent `make check` passed at this HEAD; current-head CI is green and there are no unresolved inline threads. The fix now uses the same natural widths for wrapping and tablewriter, preserving later values without an unbounded border.

Before final integration, merge current main 4b55359db9a109b70c8e1f7480662af8a1b4ff7a (#999) additively into this branch and run the required focused checks, `make check` and `make check-race`, capturing output and exit codes before pushing. Handle ordinary CI and report the exact new HEAD. Keep #992 held until #998 is merged; then integrate main into #992 and replace the buffered-only streaming claim with the actual streaming regressions already requested. No new agent or PR is needed.

apstndb:
Reviewed 41013aed0fc0f74e264129346cde9260d2180af1 against current main 4b55359db9a109b70c8e1f7480662af8a1b4ff7a. No actionable findings.

The additive main integration leaves the previously reviewed three-file width fix unchanged. Original independent empty/short-preview later-value regressions and streaming formatter tests passed again at this HEAD; independent `make check` also passed. All applicable current-head CI checks are successful. The lost-suffix finding is resolved: row wrapping and tablewriter now use consistent widths, including unknown/unlimited screen widths, without allocating an unbounded border.

PR #999, opened 2026-09-14
feat: add LIKE/EXCEPT selection for DUMP TABLES

### LINKED ISSUE #436: Add LIKE/EXCEPT pattern support for DUMP TABLES filtering
## Goal and current baseline

Support GoogleSQL LIKE/EXCEPT selection for DUMP TABLES while preserving exact table lists and qualified identity. Current main already supports named schemas and `tableID{Schema,Name}`. DUMP DATABASE is the existing no-selector/all-tables command; a missing DUMP TABLES selector remains an error. This replaces the original future-schema proposal.

## Accepted grammar and matching

```
DUMP TABLES LIKE <string> [, <string> ...]
DUMP TABLES EXCEPT <string> [, <string> ...]
DUMP TABLES LIKE <string> [, <string> ...] EXCEPT <string> [, <string> ...]
```

Keep ordinary `DUMP TABLES [schema.]table, ...` unchanged. Pattern and exact-list forms are separate; do not add FROM SCHEMA or mixed exact-list/LIKE grammar. Preserve quoted identifiers that happen to spell LIKE or EXCEPT as exact names. Require at least one valid GoogleSQL STRING literal after each clause and reject trailing tokens, empty lists and malformed literals before catalog/output work. An empty string literal is a real pattern, not no selector.

LIKE includes the union of matches; EXCEPT excludes any matching pattern from that set (or from all base tables when LIKE is absent). Pass decoded literals as SQL parameters, never interpolation. Use Spanner's native case-sensitive LIKE and native backslash escape behavior; do not add an unsupported ESCAPE clause or a Go regexp approximation. Document SQL-literal escaping separately from LIKE wildcards.

Match the display identity `IF(TABLE_SCHEMA='', TABLE_NAME, CONCAT(TABLE_SCHEMA,'.',TABLE_NAME))`: LIKE 'User%' selects matching default-schema names, LIKE '%.Users' selects Users in named schemas. Keep returned schema/name pairs separate throughout graph lookup/export. Do not split concatenated identities back into table IDs. Patterns intentionally match this text representation; exact-list quoting remains the way to select one precise identifier pair.

## Integration and safety

- Retain the existing full catalog/dependency graph. Resolve the selector through a second parameterized INFORMATION_SCHEMA query in the same read-only snapshot/transaction, honoring existing directed-read options; filter base tables consistently. Feed actual tableID pairs into the current planner.
- Resolve every selector before output. Zero base-table matches is a clear pre-output error; never pass nil as the resolved pattern selection, since nil means all tables in prepareDumpWithTxn. Keep no-selector and explicit-empty states distinct and add a narrow dumpModeTables/nil guard at the appropriate admission boundary.
- Preserve existing deterministic dependency ordering and selected-table semantics. Do not automatically add interleaved/FK parents: current DUMP TABLES exports selected table INSERTs only and requires an existing compatible target schema. DUMP DATABASE retains its DDL/all-table behavior.
- Keep source formatting/literals/writers, transaction scope, cancellation, exact-list validation and qualified output unchanged. Do not implement PostgreSQL DUMP (#435), ellipsis/exports, a new graph resolver, admin writes or unrelated cleanup.

## Target files and acceptance

Inspect `client_side_statement_def.go`, `statement_processing.go`, `statements_dump.go`, existing catalog/dependency helpers and tests. Add only the small selector representation/query helper required to keep exact and pattern forms explicit; reuse current literal/parser and catalog APIs.

Verify grammar for inclusion/exclusion/multiple/combined patterns, valid raw/escaped GoogleSQL strings, literal underscore/percent/backslash behavior, quoted LIKE/EXCEPT exact identifiers, invalid/trailing/empty-list inputs, and unchanged exact qualified lists. Exercise native LIKE against an emulator or equivalent actual Spanner fixture; do not claim regexp tests prove server behavior. Existing design probes already observed case-sensitive matches and rejection of ESCAPE syntax.

Use a catalog containing default Users/UserTest and Alpha.Users/Beta.Users, plus interleaved and FK-dependent tables. Assert selected qualified IDs and emitted INSERT targets/content/order, no implicit dependency inclusion, zero-match no-output/no-data-query behavior, and unchanged DUMP DATABASE/all and exact-list paths. Verify the selector/catalog/data share the intended read snapshot and cancellation/options. Tests must not touch managed services or user data.

Update statement help/user examples and regenerate README via make docs-update. Capture focused test/probe output and check exits, then make check and make check-race before each additive push; coverage floor80. One PR fixes #436. No dependency upgrades, history rewrites or release changes.

Baseline: mainda9b07688c3e2125a45237e98469fa2329a605db, design proof on2026-09-15JST. Existing named-schema/parent behavior was source- and emulator-verified; the LIKE/EXCEPT feature itself is not yet implemented.

apstndb:
Add `DUMP TABLES LIKE <string> [, ...] [EXCEPT <string> [, ...]]` and `DUMP TABLES EXCEPT <string> [, ...]` to select base tables by GoogleSQL LIKE patterns. Patterns are query parameters matched against schema-qualified display names; exact table lists and quoted identifiers named LIKE or EXCEPT remain supported.

Selection, catalog reads and data reads share one read-only transaction and directed-read options. Zero matches fail before output. Selected-table dumps retain their existing dependency behavior: omitted INTERLEAVE/FK parents are not added, and DUMP DATABASE still exports all tables.

Validation: focused parser/selector/emulator tests, `make docs-update`, `make check`, `make check-race`, and all applicable CI checks. Independent current-head review and local `make check` passed at 9902fc2867b470dfe725c9b6f2994b236c6a1583.

Fixes #436.


apstndb:
9902fc2867b470dfe725c9b6f2994b236c6a1583

apstndb:
Reviewed 9902fc2867b470dfe725c9b6f2994b236c6a1583 against main 41138139b1957a6206124d60e3c1aff970d67f5a. No actionable findings in the current diff.

Checked literal parsing and exact-name compatibility, parameterized schema/name selection, no-match failure before output, reuse of the read-only transaction and directed-read options, and preservation of existing selected-table dependency semantics. Independent parser/selector/emulator tests and `make check` passed at this HEAD. All applicable current-head CI checks passed; there are no unresolved review threads.

Issue #471, opened 2025-09-06
Extend URI loading functionality to other file-loading commands


apstndb:
## Goal and observed baseline

Extend existing URI loading to file-input commands without regressing local input behavior. Baseline inspected:593f37373372f72b91749205ca02d066a2373fa1 (2026-09-15).

The original statement that URI loading is only available for sample databases is obsolete. Current source has:

| Input | Current support | Remaining scope |
| --- | --- | --- |
| Sample database resources | embedded, file, HTTP(S), GCS;10 MiB limit | Preserve existing behavior |
| SQL scripts through SOURCE/meta-command and --file/--source | Bare local files, stdin/process substitution;100 MiB limit | #1007 adds explicit file/HTTP(S)/GCS URIs |
| Proto source and binary descriptor input | Bare local files and HTTP(S), with source-vs-binary URL classification and linked source imports | Reconcile GCS/file URI support after the shared SQL loader; preserve protobuf file/import identity |

Relevant code: `internal/mycli/cli.go` (`executeSourceFile`), `config.go` (`determineInputAndMode`), `sample_databases.go` (URI/HTTP/GCS loading), `system_variables.go` (proto loading/resolvers), `proto_descriptor_url.go`, and `filesafety`.

## Delivery order

1. Implement #1007 as one PR: bounded URI loading for SQL script inputs with existing stdin/local semantics and full parse-before-execution. Its child body contains the transport, cancellation, size and execution contract.
2. After that helper exists, inventory the remaining descriptor URI cases using actual source-compiler and binary-descriptor tests. Do not call all proto URI support missing: HTTP(S) already works. Separate byte loading from protobuf source import/file identity, and coordinate with #283's exact file-name conflict behavior before broadening that path. Keep this parent open until the residual scope is explicitly resolved.

An explicit remote URI supplied to a SQL input command is the opt-in to load and execute it. The current design does not add a separate global permission flag, confirmation prompt, private-network policy or authentication flow. Preserve the existing timeout/size checks and configured ADC use. This resolves the earlier policy proposal for the SQL phase; it does not authorize agents to execute arbitrary remote scripts during implementation.

## Boundaries and validation

Do not introduce a general loader plugin framework, recursive/relative SQL includes, resource cache, new credentials flow or unrelated output-file changes. Do not route100 MiB/nonregular local SQL inputs through the10 MiB regular-file sample loader unchanged. Keep source-compiler import semantics intact.

[...]

PR #1008, opened 2026-09-15
feat: complete SHOW QUERY PROFILE fingerprints

apstndb:
# Add SHOW QUERY PROFILE fingerprint completion

Fixes #523.

## Summary

Ctrl+T after `SHOW QUERY PROFILE ` (or a partial signed fingerprint) opens the existing fuzzy finder with fingerprints from the current database's `SPANNER_SYS.QUERY_PROFILES_TOP_HOUR`. Searching can use the one-line query-text preview; selecting inserts only the decimal `TEXT_FINGERPRINT`.

- Reuses `fzfItem{Value, Label}` and the bounded network completion path (timeout, cancellation, loading indicator).
- Same fetch/admission/decoding contract as `SHOW QUERY PROFILES` via a shared `fetchQueryProfilesTopHour` helper. Does not call the SHOW renderer or run the profiled user query.
- Newest interval wins on duplicate fingerprints; order is newest interval, then fingerprint.
- Profiles stay uncached. Database/schema cache, LastQueryCache, PLAN_NODE, and profile rendering are unchanged.
- `SHOW QUERY PROFILES` and unrelated commands stay on their existing completion paths.

## Test plan

- Context: singular command, partial/negative fingerprint, case/whitespace, cursor; plural and unrelated commands unaffected
- Candidates: fingerprint-only Value, searchable preview Label, Unicode truncation, empty preview, duplicate intervals, deterministic order
- Filter-level finder: distinctive query-text fragment returns exactly the fingerprint
- Fetch/admission: current-database route, loading indicator, no cache, no rows, errors, cancellation, no active-RW query
- Completion consistency: inserted statement parses as `SHOW QUERY PROFILE` with the selected int64
- Editor cancellation/no candidates leave the input buffer unchanged

[...]

PR #1010, opened 2026-09-15
feat: load SQL scripts from file, HTTP, and GCS URIs

### LINKED ISSUE #1007: Support bounded URI loading for SQL script inputs
## Outcome

Extend the existing SQL script inputs (`\.` and `--file` / `--source`) to accept explicit `https://`, `http://`, `gs://` and local `file://` URIs while preserving bare paths, stdin and process substitution. This is the SQL-input phase of #471, one PR. An explicit URI argument requests loading and executing that resource; do not add a separate global allow-remote flag or confirmation prompt.

## Current source and reuse

Baseline: main593f37373372f72b91749205ca02d066a2373fa1. `Cli.executeSourceFile` in `internal/mycli/cli.go` and `determineInputAndMode` in `config.go` both use `filesafety.SafeReadFile(...AllowNonRegular:true)` with the100 MiB default cap. `sample_databases.go` already provides HTTP/GCS loaders but applies a10 MiB sample limit and different local-file policy; blindly routing all SQL input through that aggregate loader would regress existing inputs. `system_variables.go` already supports HTTP(S) proto sources/binary descriptors, so do not redo that feature here.

## Contract

1. Keep the normal bare-path branch unchanged, including process substitution/nonregular readable files and current directory/file errors. Keep `--file -`, stdin fallback, --execute/--sql precedence, --file/--source alias precedence and interactive/batch selection unchanged.
2. Recognize explicit supported URI schemes without misclassifying normal Windows drive paths. Local file URIs must be decoded through URL semantics (including percent-encoded spaces); support empty authority and localhost, and reject other authorities with a clear local-file error. Unknown explicit URI schemes are errors, not implicit filesystem paths. Keep ordinary quoted local paths and URI arguments through the existing shellquote meta-command parser.
3. Use one small single-resource loading helper for SQL input, with caller context and caller-selected size/nonregular-file policy. Reuse existing HTTP/GCS implementations; adding a max-size parameter/wrapper to GCS is preferable to duplicating transport code. Keep the sample loader's10 MiB limit and client reuse behavior unchanged. Do not create a generic plugin/resolver package or a cache.
4. Apply the existing100 MiB SQL input limit to HTTP/GCS/file URI input too, checking both declared size and actual bytes read. Close readers/clients on success and failure. Use a bounded remote request context (30 seconds, or the earlier caller deadline) and preserve cancellation causes. Thread context through startup input loading where needed rather than using an unbounded background fetch.
5. Fetch the complete script and parse it using the existing buildCommands path before executing its first statement. A transport, limit or parse failure must execute zero script statements. Preserve the current ban on nested meta-command sourcing and current SOURCE-versus-batch EXIT behavior. Do not add nested/relative remote includes.
6. Use existing ADC for GCS and normal HTTP behavior; no login commands, credential export, new authentication flow, extra HTTP/private-network policy or managed-service writes are part of implementation/testing. Document that remote scripts execute against the selected connection just like local scripts, and GCS uses the user's configured credentials. Do not print fetched SQL as part of a load error.
7. Keep output destinations, tee behavior, proto descriptor loading and startup initialization-command semantics outside this PR. The existing protocol-specific source compiler/import identity behavior must not be incidentally changed by loader reuse.

## Acceptance and verification

- Local paths, filenames with spaces, stdin `-`, process substitution/FIFO, flag precedence, alias precedence and SOURCE/batch EXIT tests preserve their current results.
- HTTP(S) and file URI tests use local test servers/files: successful multiple-statement script, encoded path, non-200 response, declared over-limit response, chunked actual over-limit response, cancellation/timeout and reader cleanup. A failure before parsing/execution produces no database statement calls.
- GCS fake-server/client transport tests cover bucket/object decoding, attribute and streamed size limits, cancellation and cleanup, without user credentials or real buckets. Keep the production client route intact; do not add an interface solely for one implementation when existing transport/test seams suffice.
- Integration through actual Cli.executeSourceFile and startup input selection proves these paths use the new loader. A late syntax error executes zero preceding statements, and nested meta-command input remains rejected.
- Re-run relevant sample-loader and proto URL tests to show their existing limits/import semantics are unchanged.

Capture focused loader/SOURCE/input-selection tests to a file and check the exit code, then run `make check` and `make check-race`, capturing output and checking exits before each push. Update CLI flag/meta-command documentation with supported URIs and existing execution/size behavior. Keep the CI coverage floor80 and dependencies unchanged. Link the PR to this child issue and #471; the parent remains open until its remaining descriptor-URI scope is reconciled.

apstndb:
## Summary

Adds bounded URI loading for SQL script inputs (`\.` and `--file` / `--source`) as the SQL-input child of #471.

Explicit `https://`, `http://`, `gs://`, and local `file://` URIs load and execute the resource against the selected connection. Bare paths, stdin (`-`), process substitution, flag/alias precedence, parse-before-execute, nested SOURCE rejection, and SOURCE-versus-batch EXIT behavior are unchanged.

This is one PR for #1007 only. Descriptor-URI follow-up stays on #471.

## Contract

- One SQL-input helper with caller context and caller-selected size / non-regular-file policy.
- SQL input keeps the existing 100 MiB cap (`filesafety.DefaultMaxFileSize` + `AllowNonRegular`). It does not reuse the sample-database 10 MiB aggregate loader.
- HTTP/GCS reuse the existing transports; GCS gained a max-size parameter so the sample path still uses 10 MiB and its existing client reuse.
- Remote fetches use a 30s bound or the earlier caller deadline. Readers/clients are closed on success and failure.
- `file://` uses URL decoding, empty/localhost authority only. Unknown schemes error. Windows drive letters stay bare paths.
- GCS uses configured ADC. No allow-remote flag, confirmation prompt, cache, or nested remote includes.

## Test plan

- [x] Focused loader / SOURCE / input-selection tests (`TestLoadSQLInput*`, `TestCli_executeSourceFile*`, `TestDetermineInputAndMode_URI`, sample HTTP/GCS, proto URL tests)
- [x] `make check`
- [x] `make check-race`
- [x] `make docs-update` after `--file` help change

Fixes #1007
Related #471

<div><a href="https://cursor.com/agents/bc-b73f6b55-6bbd-507b-ac6b-7179a6c972f8?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-b73f6b55-6bbd-507b-ac6b-7179a6c972f8&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


apstndb:
Reviewed HEAD c420d27f19b55c85a1aff70bdbe05ba626dcd5d9. One blocking compatibility regression was reproduced.

**[P2] Preserve literal bare paths before URL parsing.** `loadSQLInput` parses every input with `url.Parse` before deciding whether it is a local path ([sql_input.go:98](https://github.com/apstndb/spanner-mycli/blob/c420d27f19b55c85a1aff70bdbe05ba626dcd5d9/internal/mycli/sql_input.go#L98)). An existing local file named `migration-100%.sql` now fails with `invalid URL escape "%.s"` through both `--file` and `SOURCE`; the previous direct `SafeReadFile` path accepted it. Explicit URI support must not apply URL escaping rules to ordinary filenames. Please classify the input first and parse only explicit URI inputs, retaining unknown-scheme errors and Windows drive paths. Add regression coverage for literal percent characters through both entry points, then run the required gates and push an additive fix.

Independent reproduction: `go test -short ./internal/mycli -run '^TestIndependent1010BarePercentPath$' -count=1`, exit 1, with both subtests failing at the same URL-escape error. No fetched SQL or remote service was involved.

apstndb:
b3471d2f0626424f6b79be7ac4fffa5d4d76f7ca

apstndb:
The percent-path finding is fixed in HEAD b3471d2f0626424f6b79be7ac4fffa5d4d76f7ca. The original independent SOURCE/--file reproduction now passes, together with the focused loader, alias and SOURCE tests (exit 0). No further actionable finding in the feature diff and the inspected fix.

Main has since advanced to d6e28c2549fa5b54bcbf11fc6dde566efe4b8a40 (#1008). Please merge current main additively, preserve both generated README changes, and rerun `make check`, `make check-race` and current-head CI. Keep the PR draft for final exact-head acceptance.

apstndb:
817d82e9b3a53425ae283589ccf2658d30998e5b

apstndb:
No actionable findings remain in the reviewed scope at HEAD `817d82e9b3a53425ae283589ccf2658d30998e5b`, based on current main `d6e28c2549fa5b54bcbf11fc6dde566efe4b8a40`.

The literal-percent local path regression is fixed: the original independent `--file` and `SOURCE` reproduction now succeeds. I reviewed the URI/local-path classification, bounded and cancellable loading, execution integration and the subsequent main-branch merge. Independent `make check` passes, including that extra regression test; all applicable current-HEAD CI checks, including race, pass. There are no unresolved review threads.

The first local gate stopped only on formatting of my untracked reproduction test; formatting that test and rerunning the gate succeeded. This did not require a PR source change. Windows compilation is covered by CI; I have not runtime-tested Windows subprocess behavior.

This completes the SQL-input phase in #1007; #471 retains the separate descriptor URI scope.

PR #1011, opened 2026-09-15
feat: add PULL REMOTE PROTO for local descriptor install

apstndb:
## Summary

Implements `PULL REMOTE PROTO` so remote PROTO/ENUM descriptors can be loaded into the local session store for decoding and `SYNC PROTO BUNDLE` composition.

Fixes #283.

- Syntax: `PULL REMOTE PROTO ALL`, a single GoogleSQL type name, or a parenthesized list. First-occurrence dedup; empty/malformed/trailing input is rejected.
- Fresh Admin `GetDatabaseDdl` read (`GetDatabaseDdlFresh`), never the 30s SHOW/DDL cache. No remote DDL.
- File-granularity install plus import closure. Prefer remote files; reuse matching local imports. Do not invent fields or fetch extra network/filesystem dependencies. Unrelated broken remote files do not block a complete selected component.
- Same-file complete updates are allowed when the merged candidate validates. Untouched local files survive. Removal or placeholder downgrade of complete local types is rejected. The same full name in different file names is a conflict.
- `protodesc.NewFiles` validates the whole candidate before commit. Errors leave `Internal.ProtoDescriptor` and `Internal.ProtoDescriptorFile` unchanged. Graph changes clear file paths; no-op pulls keep them.
- Result uses the `SHOW LOCAL PROTO` row shape for selected roots only. Detached sessions and open manual batches are rejected. Existing descriptor-variable transaction rules are unchanged.

## Out of scope

- No managed-service writes or completeness claim.
- No change to generic `mergeFDS` behavior.
- No TypeName-walking import invention, dependency upgrades, or descriptor-rewriting framework.

## Verification

Go 1.26.8, real Docker. Focused parser/planner/Execute tests, existing proto/SYNC tests, and emulator integration (`TestPullRemoteProtoImportedBundle`, `TestPullRemoteProtoRootOnlyMissingImport`) plus `make docs-update`, `make check`, and `make check-race` before push.

[...]

PR #1012, opened 2026-09-15
feat: retry ABORTED explicit read-write transactions (#1006)

### LINKED ISSUE #1006: Enable bounded explicit ABORTED retry with observation-aware prefix replay
## Outcome and dependency

Complete the explicit transaction and observation-completion phases of #293 after shared capture prerequisite #1005 and merged #994/#997. With RETRY_ABORTS_INTERNALLY=TRUE, an explicit read-write transaction can recover from ABORTED by replaying its fully observed prefix and retrying the current operation without printing earlier results twice. A changed replay result terminates the logical transaction.

Keep the default FALSE and document the difference from the Java/Go driver defaults. Do not ship explicit replay separately from its output-completion guard: those two parts jointly make this user-facing feature correct.

## Entry points and state

Inspect `transaction_manager.go`, `transaction_manager_auto_dml.go`, `retry_aborts.go`, `session_transaction_context.go`, `savepoint_capture.go`, `savepoint_journal.go`, `savepoint_replay.go`, `savepoint_receipt.go`, `execute_sql.go`, `execute_dml.go`, `statements_explain_describe.go`, `statements_system_variable.go`, and the existing row/result completion paths.

- Remove phase-1 rejection of explicit/pending RW owners with captured TRUE only after all paths below are implemented. Preserve session SET/SHOW/RESET semantics: SHOW is the session value; an existing owner keeps its captured value after later session changes.
- Allow SET LOCAL RETRY_ABORTS_INTERNALLY=TRUE/FALSE only on an otherwise unused pending RW owner, before first RPC, queued operation or capture/marker, with no operation in flight. Update that owner and exactly-once LOCAL undo consistently. A later ordinary SET overrides undo for that name, as existing SET LOCAL does. Reject changes after use before changing registry/owner state. Direct BEGIN RW has already acquired a physical transaction and has no such window; implicit RW has no LOCAL window. Follow the existing rejection convention for RO/no-owner cases.
- Journal from BEGIN when the owner needs retry, including operations before its first SAVEPOINT and after its last marker is released. No second journal or SQL reparse/parameter reevaluation during retry.

## Retry contract

1. Retry only a real Spanner ABORTED outcome on current SQL (including SELECT inside RW), DML/THEN RETURN, PROFILE execution, automatic/manual Batch DML, or Commit. Buffered MUTATE entries must be replayed on the replacement physical attempt. PLAN-only operations, PDML, read-only transactions, validation errors and ambiguous/non-ABORTED Commit failures keep their existing behavior.
2. Stop/discard the failed physical attempt while preserving the logical owner, LOCAL undo, original TRANSACTION_TIMEOUT/caller budget, frozen constructor/options/tags and admitted-work history. Reuse #997's SDK retry-delay extraction and interruptible bounded wait. Physical reconstruction and replay are not new idle user activity.
3. Reconstruct using frozen options, replay the successful prefix silently through the existing typed ordered fingerprint helpers, then retry the current frozen operation. Replay compares type/metadata, row values/order/count and affected-row or per-statement batch counts, including zero rows. Do not fingerprint formatted strings, re-render prior rows, update LastQueryCache for silent prefix operations, or replay a failed/unobserved receipt as successful.
4. Prior fully completed output is allowed: replay it silently and require its receipt to match. Bytes already delivered by the CURRENT failed operation, including headers/presentation bytes, prohibit automatic re-execution of that operation. Buffered/unpublished current results may be discarded and retried. A writer failure is not a successful receipt or a retryable database failure. This distinction must hold for table/vertical streaming, buffered query results, DML returned rows and PROFILE plan/stats output.
5. Keep at most 49 automatic physical reconstructions (50 attempts including the original) per logical owner. Repeated ABORTED errors during prefix replay consume the same budget; successful intervening statements and manual ROLLBACK TO do not reset it. Keep attempt identity monotonic and distinguish it from the bounded retry counter if needed. Respect the earliest original caller/owner deadline throughout; do not reset the transaction timeout or mislabel caller expiry as owner expiry.
6. When retry is enabled and eligible, try full-prefix automatic recovery before entering the existing SAVEPOINT recovery-required state. Successful retry preserves marker positions and LOCAL undo. If retry is forbidden by partial current output or exhausts its ABORTED attempt budget, use the existing valid-marker recovery behavior where the logical owner remains usable; without a marker, retire it. Expired/cancelled logical owners and fingerprint mismatch/reconstruction failure are terminal and must retire the owner and restore LOCAL exactly once. Manual SAVEPOINT recovery must never silently substitute rollback-to-an-earlier-marker for full-prefix retry.
7. Cancellation, non-ABORTED replay failure, mismatch, journal-limit admission failure and stale completion must return the appropriate cause without extra writes, duplicate output or an endlessly retried owner. Preserve #964 SQL-versus-Commit fallback eligibility and #989 owner/attempt completion guards.

## Required evidence

- Actual SQL/SDK path: BEGIN with retry enabled and no marker; ABORTED on first SELECT/DML; abort after fully displayed reads; Commit abort; automatic/manual Batch DML; buffered MUTATE; THEN RETURN and PROFILE. Assert successful final results and exact RPC/effect counts, not merely any error or nil error.
- Typed mismatch cases: changed type/value, row order/count, zero rows and DML/batch counts. A changed prior observation prevents Commit and retires the logical transaction.
- Prior completed output appears once; current partial output prevents retry; unpublished buffered output can retry; output failure never supplies a successful fingerprint. Cover both real streaming and buffered routes.
- No-marker retry works with SAVEPOINT commands disabled. Enabled markers survive successful retry; marker release retains retry history; exhausted/partial-output failures preserve only existing valid recovery behavior; manual ROLLBACK TO does not reset the automatic budget.
- Session SET/RESET versus captured owner, permitted/rejected SET LOCAL transitions, SET-after-LOCAL undo, original deadline causes, cancellation during delay/replay and stale owner/attempt completions.
- Existing default-FALSE and implicit #997 behavior remains unchanged. Preserve PLAN/PDML/RO exclusions and non-ABORTED Commit behavior.

## Delivery

One implementation PR after the capture prerequisite. Update `docs/savepoint.md`, system-variable reference and driver-compatibility matrix with the opt-in behavior, lifetime budget and hidden-read/unreturned-volatile-write limitations; regenerate reference metadata with `make docs-update`. Run selected new/existing RPC and SAVEPOINT tests, then `make check` and `make check-race`; capture every output and check exit codes before each push. The CI coverage floor stays80. No dependency upgrade, default-TRUE change, general renderer or separate replay engine.

apstndb:
## Summary

Implements #1006: with `RETRY_ABORTS_INTERNALLY=TRUE`, an explicit read-write owner recovers from a real Spanner `ABORTED` by silently replaying the fully observed journal prefix and retrying the current frozen operation. A changed replay result or reconstruction failure ends the logical transaction.

This ships replay and the output-completion guard together. Default remains `FALSE` (Java/Go drivers default `TRUE`).

## Behavior

- Journal from `BEGIN` when the owner needs retry, including work before the first `SAVEPOINT` and after the last marker is released. No second journal or SQL reparse during retry.
- Retry real `ABORTED` on SQL (including `SELECT` in RW), DML/`THEN RETURN`, PROFILE, automatic/manual Batch DML, and Commit. Buffered `MUTATE` is replayed. PLAN-only, PDML, RO, validation errors, writer failures, and non-`ABORTED` Commit keep existing behavior.
- Discard the failed physical attempt; keep the logical owner, `SET LOCAL` undo, original `TRANSACTION_TIMEOUT`/caller budget, frozen constructor/options/tags, and admitted-work history. Reuse #997 delay extraction. Reconstruction is not new user-idle activity.
- Prior completed output is replayed silently and must match. Bytes already delivered by the **current** failed operation (including headers, or a streaming row/header already presented to the sink) forbid re-execution. Unpublished buffered results may be discarded and retried.
- Lifetime budget: 49 automatic reconstructions (50 attempts including the original). Prefix-replay `ABORTED` consumes the same budget. Successful statements and manual `ROLLBACK TO` do not reset it.
- Full-prefix automatic recovery runs before SAVEPOINT recovery-required. Partial output or exhausted budget keep existing valid-marker recovery; without a marker the owner is retired. Fingerprint mismatch, reconstruction failure, cancellation, and owner expiry are terminal.
- `SET LOCAL RETRY_ABORTS_INTERNALLY` is allowed only on an unused pending RW owner, before first RPC, queued work, or marker. Direct `BEGIN RW` and implicit RW have no LOCAL window. Session `SET`/`SHOW`/`RESET` stay session-scoped; a live owner keeps its captured value.

## Evidence

RPC tests in `retry_aborts_explicit_rpc_test.go` cover first SELECT/DML abort, Commit + MUTATE, manual/automatic Batch DML, THEN RETURN, PROFILE, typed fingerprint mismatches (value/type/order/count/zero rows/DML/batch), buffered unpublished retry, prior output once, marker survival/release, `ROLLBACK TO` not resetting the budget, exhausted-budget marker recovery, SET LOCAL window, session SET vs captured owner, stale completion, cancel during delay/replay, default FALSE, PLAN exclusion, writer failure, owner deadline, and non-ABORTED Commit.

## Docs

Updated `docs/savepoint.md`, `docs/system_variables.md` (`make docs-update`), and `docs/spanner-driver-compatibility.md` with opt-in behavior, lifetime budget, LOCAL window, and hidden-read/unreturned-volatile-write limitations.

Fixes #1006
Related #293

<div><a href="https://cursor.com/agents/bc-bae5a297-62a2-5f7c-8de5-bdef2c15733d?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-bae5a297-62a2-5f7c-8de5-bdef2c15733d&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


apstndb:
Review of HEAD `a49e49300540f07cadb89c09687d011a1e4ff2bb`: two actionable findings reproduced so far. This is not final acceptance.

1. **[P1, runtime-reproduced] Bind query retry to the admitted owner and attempt.** `explicit_abort_retry.go:268` accepts only the error and output flag, so the new calls in `execute_sql.go:389,421` ignore the capture token and single-use runner boundary. With retry-enabled BEGIN RW, an ABORTED from `session.ExecuteStatement(..., &ExportDataStatement{...})` now reconstructs that unrelated RW transaction and executes the export again; the command returns nil instead of the original ABORTED. The independent SDK test observes BeginTransaction count1->2 and a different physical handle. A separate delayed-completion injection using an old token after ROLLBACK/BEGIN likewise retries the old SQL in the replacement owner (runner calls2, begins2->3), despite existing owner/attempt failure guards. Require an eligible admitted operation belonging to the still-current owner AND attempt before any reconstruction/fallback; preserve the original error and leave the replacement/isolated owner untouched for nil, stale, single-use and excluded PLAN/RO paths. Apply the same binding to PROFILE. Checking only the session's current retry flag is insufficient.

2. **[P2, runtime-reproduced] Freeze options for the current SELECT/PROFILE operation across retries.** `execute_sql.go:385` calls the ordinary runner again, which rebuilds options and consumes STATEMENT_TAG each time. With `SET STATEMENT_TAG = 'keep-on-retry'`, the first actual SELECT RPC has that tag but the successful second attempt has an empty tag. The journal's frozen successful prefix does not freeze the currently failing operation. Reuse the initially resolved statement/options/query mode for all attempts, consuming the session tag once; do not reread mutable query settings on retry. Check the equivalent PROFILE loop in `statements_explain_describe.go` too.

Independent reproduction command: `GOTOOLCHAIN=go1.26.8 go test -short ./internal/mycli -run '^TestIndependent1012' -count=1` (exit1, three failing cases). The tests use the existing local SDK RPC harness; no managed database or export was executed. The stale-token case deliberately injects a delayed ABORTED into the production runner completion branch. Add regressions asserting the original error cause, physical owner/attempt, RPC counts and retry request options, not only a boolean helper result.

Please fix in this PR, integrate current main additively, then run focused tests, `make check`, `make check-race` and applicable CI with captured exit codes. Keep the draft; no force-push or merge. Main now includes #1011 at `22d7a7e47df8c8592612878aaa458a6bb1130d84`.

apstndb:
Resuming this deferred PR after v0.35.0. Review of HEAD `7b43b2d379a873cb21cee0662fd9475cd23e9ec4`: the earlier single-use/stale query-owner and statement-tag findings pass the original independent tests. Two additional findings are runtime-reproduced; this is not final acceptance.

1. **[P1] Preserve the replacement owner when a Commit retry loses its original owner.** [transaction_manager.go:1041](https://github.com/apstndb/spanner-mycli/blob/7b43b2d379a873cb21cee0662fd9475cd23e9ec4/internal/mycli/transaction_manager.go#L1041) retires whatever `tm.tc` exists after `recoverExplicitAbortLocked` returns. Recovery unlocks `tm.mu` during backoff. An independent SDK test makes the first actual Commit return ABORTED, then uses the existing retry-wait seam to call `RollbackReadWriteTransaction` followed by `BeginReadWriteTransaction`. Recovery correctly returns `errImplicitAbortRetryLostOwner`, but the outer Commit completion then clears the newly created transaction (`tm.tc == nil`). Bind terminal retirement and recovery decisions to the owner/attempt that issued this Commit; a delayed completion must leave the replacement and its LOCAL/idle state untouched. Add a regression through the actual Commit path, including the owner change during the unlocked wait.

2. **[P2] Distinguish consumed PROFILE rows from published output.** [statements_explain_describe.go:497](https://github.com/apstndb/spanner-mycli/blob/7b43b2d379a873cb21cee0662fd9475cd23e9ec4/internal/mycli/statements_explain_describe.go#L497) passes `actualRows > 0` as current-output-delivered. EXPLAIN ANALYZE has not rendered any result at that point. A real SDK stream emitting one complete row with a resume token, then ABORTED, returns ABORTED with only one SQL RPC instead of discarding the unpublished attempt and retrying successfully. Use actual presentation completion for this path; internal row observation alone cannot prohibit retry. The existing midstream fixture lacks resume tokens, so the SDK can buffer all rows and those tests miss this branch. Add a resumable-row fixture and verify final two-row PROFILE result plus exactly two SQL RPCs. Keep actual partial CSV/vertical output non-retryable.

Independent commands (local SDK RPC harness; no managed writes):
- `GOTOOLCHAIN=go1.26.8 go test -short ./internal/mycli -run '^TestIndependent1012' -count=1`: the original three tests pass before adding these probes.
- `GOTOOLCHAIN=go1.26.8 go test -short ./internal/mycli -run '^TestIndependent1012(Profile|Actual)' -count=1 -v`: PROFILE fails as above; actual partial CSV and vertical presentation correctly stop after one RPC, and buffered TABLE retries successfully.
- `GOTOOLCHAIN=go1.26.8 go test -short ./internal/mycli -run '^TestIndependent1012Commit' -count=1 -v`: replacement-owner test fails; ordinary Commit after automatic-flush retry passes.

Please resume this same PR, add scoped fixes/regressions, integrate current main additively if it moves, and capture focused tests, `make check`, `make check-race` and applicable CI at the new HEAD. Keep the draft until acceptance; no force-push or merge.

apstndb:
Reviewed HEAD 256bb5a5ae8b4f8665fbb8ed90f7847ad0be2895. The findings in both prior reviews are addressed, and there are no remaining actionable findings in the reviewed diff. Independent tests now pass for single-use/stale owners, frozen request options, Commit owner replacement, unpublished PROFILE rows, actual partial CSV/vertical output, buffered TABLE retry, and automatic DML flush followed by Commit. Independent make check also passed (exit 0), as did the current required CI checks.

Integration is pending: #1013 has just merged as f7d616a923ec3173b126cfa976d11cbb070ad474, so this PR is now behind main. Please merge current origin/main additively, preserve SHOW HISTORY in the recovery-safe documentation, run the required validation and own CI through completion. Keep the same PR and report the updated HEAD. No force-push or merge; final acceptance will refresh the resulting diff and checks.

apstndb:
Final review of HEAD 58bf2d96efc5531be68b9f68b7f3b642fe274bbc: no actionable findings. The integration tree exactly matches an ordinary merge of the independently reviewed retry implementation (256bb5a5) and SHOW HISTORY main (f7d616a9), with no additional implementation changes. SHOW HISTORY remains documented as recovery-safe.

Independent regression and SHOW HISTORY tests passed on this head (exit 0). The retry implementation's independent make check passed, and all current required CI checks, including race and cross-compilation, passed at this integrated head. No unresolved threads or pending reviews remain. The two earlier review rounds are addressed; ready to merge within this reviewed scope.

PR #1014, opened 2026-09-15
feat: load proto descriptors from file:// and gs:// URIs

### LINKED ISSUE #471: Extend URI loading functionality to other file-loading commands

apstndb:
Load protobuf descriptors from file:// and gs:// URIs through --proto-descriptor-file and SET/ADD PROTO_DESCRIPTORS_FILE_PATH, for both source .proto and binary descriptor sets.

File URIs share the decoded local path's identity, including mixed imports in one source graph, and use the regular-file/100 MiB policy. GCS uses application default credentials and keeps explicit URI identity; relative imports are not rewritten onto a bucket. Startup cancellation reaches the loader, remote fetches are bounded, and failed loading or compilation preserves the installed graph.

Validation covers local source/binary aliases, mixed imports, oversized/non-regular inputs, fake GCS reads, cancellation and atomic failure handling, plus existing HTTP behavior. Current-head validation and review receipts are recorded in the review comments.

Fixes #471.


apstndb:
Review of HEAD fad32dc9fabf4b27be1dd7026a3f202a384e6c47: one blocking finding, runtime-reproduced.

**[P2] Apply the local file policy before compiling a file:// source root.** [system_variables.go:750](https://github.com/apstndb/spanner-mycli/blob/fad32dc9fabf4b27be1dd7026a3f202a384e6c47/internal/mycli/system_variables.go#L750) decodes the URI into a bare path. The first resolver ([line 599](https://github.com/apstndb/spanner-mycli/blob/fad32dc9fabf4b27be1dd7026a3f202a384e6c47/internal/mycli/system_variables.go#L599)) then opens that path directly, so the SafeReadFile branch never runs for source roots. A sparse .proto file with size DefaultMaxFileSize+1 and initial byte ! reaches the parser and returns invalid character instead of a size rejection. Thus the advertised 100 MiB/regular-file policy applies to binary roots and explicit file imports but not file:// source roots; special-file sources can also bypass the pre-open check (source-backed consequence, not separately runtime-probed).

Please use the existing safe loader for local source resolution while preserving the decoded descriptor identity, standard imports, bare percent/Windows paths and atomic install. Cover oversized and non-regular source roots without hanging the suite, and retain valid source/binary alias tests. Keep this same PR, run focused tests, make check and make check-race, and own CI/base integration as before. Do not force-push or merge.

Also rename or simplify the test titled "client closed after success and failure": it counts client creation, not Close calls, so its name currently claims evidence it does not collect. There is no need to add a production abstraction just to observe Close; the production defer is directly inspectable.

apstndb:
One additional runtime-reproduced issue at the same reviewed HEAD fad32dc9fabf4b27be1dd7026a3f202a384e6c47, while checking the promised decoded-path identity.

**[P2] Canonicalize file URI import identity before linking the source graph.** [system_variables.go:781](https://github.com/apstndb/spanner-mycli/blob/fad32dc9fabf4b27be1dd7026a3f202a384e6c47/internal/mycli/system_variables.go#L781) canonicalizes the resulting descriptors only after compilation succeeds. With root.proto importing a.proto and b.proto, a.proto importing /absolute/dep.proto and b.proto importing file:///absolute/dep.proto, both branches refer to the same file but compilation fails first: symbol "mixed.Dep" already defined at file:///.../dep.proto. The promised file URI alias therefore cannot be used in a mixed source graph, even though the two individual inputs work separately.

Reproducer: dep.proto defines package mixed/message Dep; a.proto imports its absolute path and defines A{Dep d=1}; b.proto imports its file URI and defines B{Dep d=1}; root imports a and b and defines Root{A a=1; B b=2}. Compiling the root through its file URI fails with duplicate symbol. Command: go test -short ./internal/mycli -run '^TestIndependent1014Mixed' -count=1 (exit1).

Please fix this in the same lane alongside the size policy, using the compiler's existing resolver/descriptor facilities and a narrow mixed-graph regression. Preserve bare-path/file-URI shared identity and GCS URI identity; do not add a generic loader framework or modify external dependencies. If a correct bounded alias implementation requires a material design change, report the specific constraint and options rather than silently weakening the alias claim. The other authority and validation boundaries remain unchanged.

apstndb:
Review of HEAD 0e232bfed83e7ccc76d6cfd1b509ebc59ab11bde: both findings are addressed, with no remaining actionable semantic findings. Independent tests now verify that oversized file URI sources are rejected before parsing, and that mixed bare-path/file-URI imports compile into one usable descriptor graph. The related local, HTTP and fake-GCS tests also pass (exit 0). Source review confirms per-load context, bounded remote transport and atomic candidate installation.

#1015 has now merged as b19197653365f41e3aa348c294438b84e4e6bdf8. This is the final selected lane: please merge current origin/main additively, run make check/make check-race and own CI to completion, then return the new HEAD. GitHub requires an up-to-date branch. Preserve MAX_PARTITIONS, retry behavior and history; no further feature work, force-push, auto-merge, admin bypass or merge. Final acceptance will verify this integration delta and current-head gates.

apstndb:
Final review of HEAD 53abb41feb4fb7e3d3fd6bbe81ad140579461ff1: no actionable findings. Both earlier findings are fixed. Its tree b15dc7c0ed0a9cea84d416421089967154919f75 exactly equals the ordinary merge of reviewed descriptor head0e232bfe and mainb1919765, with no additional implementation changes.

The independent oversized-source and mixed-identity regressions pass, along with related local/HTTP/fake-GCS tests, make check and the integrated descriptor/MAX_PARTITIONS tests (exit 0). All required current-head CI checks passed, including race and cross-compilation. No unresolved threads or pending reviews remain. Ready to merge within the reviewed scope; GCS behavior was verified against the fake transport, not a managed bucket.

PR #1016, opened 2026-09-15
Advertise SET LOCAL and RESET in HELP VARIABLES operations

apstndb:
The Operations column in `HELP VARIABLES` and the generated system-variable reference now includes `local` (`SET LOCAL`) and `reset` (`RESET` / `RESET ALL`). These labels come from existing registry policy and describe supported operations; transaction phase, setter guards, valid values and startup snapshots still determine whether a particular request succeeds.

The three-column layout, canonical variable names, descriptions and alias exclusion are preserved. Unimplemented variables retain their single marker. Rendering this metadata does not read live values or invoke setters. The shared information type also removes repeated anonymous struct declarations.

Related to #292 and #725. Type/default metadata and the broader documentation work remain separate.

Validation at `6a4fb954b6704e21e3367c85465f8f3effff12bb`: focused HELP/registry tests and independent `make check` passed with Docker/emulator available. Required CI, including race, cross-compilation and generated-document synchronization, passed. The initial implementation VM could run only short tests/lint/format checks because Docker was unavailable; the full local gate was subsequently verified independently before merge.


apstndb:
Reviewed `6a4fb954b6704e21e3367c85465f8f3effff12bb` against `2a662443c6355cdb8e7ba9fa99da9460ac54fa68`: **no actionable findings** in the operation-metadata change.

The labels follow the existing LOCAL/RESET policies, preserve the unimplemented marker and feature/alias behavior, and do not call Get/Set. All 117 canonical documentation rows and their descriptions are unchanged apart from Operations cells. The prose correctly distinguishes static support from permission in the current transaction phase.

Focused HELP/registry tests and independent `make check` (including emulator integration tests) passed at this head. Required CI also passed, including race and generated-document synchronization. No runtime SET/LOCAL/RESET behavior change is included.

